PR #61, opened 2021-02-16
Safe 117 confirm rerouting while in form edition

adrianmujica153:


projects/back-office/src/app/dashboard/pages/form-builder/form-builder.component.ts
```diff
@@ -42,13 +45,36 @@ export class FormBuilderComponent implements OnInit {
...
     private router: Router,
-    private snackBar: WhoSnackBarService
-  ) { }
+    private snackBar: WhoSnackBarService,
+    public dialog: MatDialog,
+  ) {
+
+    /* Shows modal confirmation before refresh or close the page if has changes on form
+    */
+    window.addEventListener('beforeunload', (event) => {
```
AntoineRelief, line 61: there is one issue with that, if the user wants to logout from a modified form, because the logout action will be triggered before the confirmation, so the user can stay on the form page but unauthenticated

AntoineRelief (changes requested):
Check comment about logout

AntoineRelief (approved):

PR #115, opened 2021-03-31
Safe 232

PacomeRivier:
I did move the workflow service to the common library and use it to pass data from one step to another. I don't know if it's a good practice.


PacomeRivier:
I also put a fix for the token renewal in the first commit !

PR #127, opened 2021-04-13
Safe 248 gerenate word excel from selected rows

AntoineRelief:


projects/who-ems/src/lib/components/widgets/grid/grid.component.html
```diff
@@ -31,6 +31,9 @@
         <button *ngIf="(!settings.actions || settings.actions.convert)" class='k-button'
           (click)="onConvertRecord(selectedRowsIndex)" style="float: right" [disabled]="!canUpdateSelectedRows">Convert
         </button>
+        <button *ngIf="(!settings.actions || settings.actions.exportMultiple)" class='k-button'
```
AntoineRelief: no need to have an action for that, enable it per default

projects/who-ems/src/lib/components/widgets/grid/grid.component.ts
```diff
@@ -680,6 +680,36 @@ export class WhoGridComponent implements OnInit, OnChanges, OnDestroy {
     });
   }
 
+  /* Open a confirmation modal and then delete the selected record
+  */
+  public onExportRecord(items: number[]): void {
```
AntoineRelief: you can look at how we use downloadService for that ( ex in form-records.component ).
The back-end API for that doesn't allow listing ids to export for the moment, that I think we can expand it in order to make it
It would result in options to pass to the http.get call ( front-end) and filtering in the back-end based on request parameters
if you look at router.get('/form/records/:id', async (req, res), for the moment we try to get all Records attached to the detected form, but we could filter based on list of ids

PR #158, opened 2021-05-04
[SAFE-271] add the sidenav for history

AntoineRelief:


projects/back-office/src/app/dashboard/pages/form-records/form-records.component.ts
```diff
@@ -79,6 +85,27 @@ export class FormRecordsComponent implements OnInit {
...
+      variables: {
+        id
+      }
+    }).subscribe(res => {
+      this.layoutService.setRightSidenav({
+        factory: this.factory,
+        inputs: {
+          record: res.data.record,
+          revert: (item: any, dialog: any) => {
+            console.log('action !');
```
AntoineRelief: @Hugo-Plissonneau we should replace that with the confirmRevertDialog that we also have in the widget grid component at the same place

PR #174, opened 2021-05-13
Safe 286

adrianmujica153:


package.json
```diff
@@ -53,12 +53,15 @@
     "@progress/kendo-drawing": "^1.5.8",
     "@progress/kendo-recurrence": "^1.0.0",
     "@progress/kendo-theme-default": "latest",
+    "@types/lodash-es": "^4.17.4",
```
AntoineRelief: ⚠️ @adrianmujica153 For imports of types, use the --save-dev flag with npm i, so it's put under devDependencies ( no need to have them for builds )

projects/back-office/src/app/dashboard/pages/applications/applications.component.scss
```diff
@@ -23,3 +23,10 @@
     margin-left: 12px;
   }
 }
+
+:host {
```
AntoineRelief: ⚠️ @adrianmujica153 put at the top of file :host

projects/safe/src/lib/components/layout/layout.component.scss
```diff
@@ -9,7 +9,6 @@
 }
 
 .sidenav-content {
-    min-height: calc(100% - 64px);
```
AntoineRelief, line 12: @adrianmujica153 is that needed for the MR ?

adrianmujica153: I think that is not necessary.

PR #179, opened 2021-05-19
[SAFE-318] button to switch from simple to advanced filter grid

AntoineRelief:


projects/safe/src/lib/components/widgets/grid/grid.component.ts
```diff
@@ -72,6 +71,27 @@ const MULTISELECT_TYPES: string[] = ['checkbox', 'tagbox'];
 */
 export class SafeGridComponent implements OnInit, OnChanges, OnDestroy {
 
+  get hasChanges(): boolean {
+    return this.updatedItems.length > 0;
+  }
+
+  constructor(
```
AntoineRelief: ⚠️ @Hugo-Plissonneau constructor should always be declared after the public / private / Input / Output

PR #180, opened 2021-05-20
[SAFE-303] Update default grid headers

AntoineRelief:


projects/safe/src/lib/components/widgets/grid/grid.component.ts
```diff
@@ -946,6 +946,14 @@ export class SafeGridComponent implements OnInit, OnChanges, OnDestroy {
     return ( e.offsetWidth < e.scrollWidth );
   }
 
+  /* Pretify grid default title
+  */
+  public transformGridTitle(myValue: string): string {
+      myValue = myValue.replace('_', ' ');
```
AntoineRelief: ⚠️ @TristanLeGoff myValue is not a suitable name in dev, for test that's fine but not for deployed code. Better to use names that reflect what the variable is about

PR #202, opened 2021-06-07
Safe 374

AntoineRelief:


projects/safe/src/lib/components/import-record-modal/import-record-modal.component.css
```diff
@@ -0,0 +1,7 @@
+#dropzoneCont {
```
AntoineRelief: @MartinGbz for css try to prefer classes and snake case for names

PR #289, opened 2021-09-20
AB#10627

AntoineRelief:
# Description

A grid can now load more than 25 records, it will load the records regarding as how many items the user wants to display on the grid.
It does generate a new error, NG0100 linked to the "isEllipsisActive" changing from true to false when loading the new records, but I don't know how to fix it.

## Type of change

# How Has This Been Tested?

I tested it with a form with 78 records and it works fine.

- [x] Test A

# Checklist:

### DIFF
--- a/projects/safe/src/lib/services/query-builder.service.ts
+++ b/projects/safe/src/lib/services/query-builder.service.ts
@@ -129,6 +129,7 @@ export class QueryBuilderService {
   }
 
   public buildQuery(settings: any, skip?: any, take?: any): any {
+    // adapter tous les res de buildQuery pour mettre res.result resource grid component, map component, grid component,
     const builtQuery = settings.query;
     if (builtQuery && builtQuery.fields.length > 0) {
       const fields = ['canUpdate\ncanDelete\n'].concat(this.buildFields(builtQuery.fields));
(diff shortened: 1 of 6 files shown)


projects/safe/src/lib/services/query-builder.service.ts
```diff
@@ -128,17 +128,18 @@ export class QueryBuilderService {
...
       const fields = ['canUpdate\ncanDelete\n'].concat(this.buildFields(builtQuery.fields));
-      const metaFields = this.buildMetaFields(builtQuery.fields);
       const query = gql`
         query GetCustomQuery {
           ${builtQuery.name}(
           sortField: ${builtQuery.sort && builtQuery.sort.field ? `"${builtQuery.sort.field}"` : null},
           sortOrder: "${builtQuery.sort?.order || '' }",
-          filter: ${this.objToString(this.buildFilter(builtQuery.filter))}
+          filter: ${this.objToString(this.buildFilter(builtQuery.filter))},
+          perPage: ${take ? take*3 : 25},
```
AntoineRelief: @Hugo-Plissonneau I think you can keep the 25 there, the update should be somewhere else
I do think the skip/take operation can be an issue, so keep the skip ? skip : 0

PR #292, opened 2021-09-22
AB#10625

MartinGbz:
# Description

The restrictions on the points specified in the ticket are implemented.

But there 4 informations to know:

- In the FO part: it was specified to add the canSee Page Permission in the condition which enable us to see the form in the page. It has been done, but if we don't have the canSee Page permission, the page isn't displayed in the list of form/dashboard/workflow of the application (and we can't access it through the url, so this it's well protected). But therefore test the canSee in order to display the form in the page is unnecessary (because we can't see the page which contains this form). I let it just in case (@PacomeRivier advise me to let it too).

- In the BO part (in the first point):
Same things than the FO part:
     - when canSeeApplications & canManageAapplications are both false, you can't see the list of application (so I think put restriction on the form in the form page is useless but I let it).
     - when only the canManageAapplications is false, all the applications are displayed but all these form/dashboard/workflow aren't display (so another time it seems also useless to put restriction but I let it).

- In the BO part (in the second point):
To test this, I verified the canManageForms of the user and the canUpdate permissions of the form.
But when I was trying to test my changes, I didn't succeed to have the canUpdate permission of the form to false when canManageForms was true. (I tried to create a role without form updating rights, but as soon as canManageForms value was true, the canUpdate was always true).

We should find a way to test the canUpdate condition only.

- Finally there is a weird behavior at the line 120 of the form.component file in the BO (I put a comment)

## Type of change

# How Has This Been Tested?

- [x] Test A: FO part

Try to set the form in pending and active (pending => form not displayed, active => form display).
Do the same things for the canSee permission of the page which contains this form (fasle => not display, true => display).
Do the same things for the both at the same time (false and true).

- [x] Test B: BO part (first point)

Try to set canSeeApplications and canManageAapplications to false and true in all different cases (following the same logic as the previous point).

- [x] Test B: BO part (second point)

Same than the previous point

# Checklist:


projects/back-office/src/app/application/pages/form/form.component.html
```diff
@@ -1,30 +1,35 @@
 <mat-spinner diameter="45" class="page-loader" *ngIf="loading"></mat-spinner>
-<ng-container *ngIf="!loading && form">
+<ng-container *ngIf="user.permissions">
+  <ng-container *ngIf="!loading && form && user.permissions.some(isUserHasCanSeeAppPermission) && user.permissions.some(isUserHasCanManageAppPermission)">
```
AntoineRelief: @MartinGbz ⚠️ for future PRs, better to have a variable set for that, than to put complex conditions in the html directly

projects/front-office/src/app/dashboard/pages/form/form.component.ts
```diff
@@ -5,6 +5,8 @@ import { ActivatedRoute, Router } from '@angular/router';
 import { Form, Page, Step, SafeFormComponent } from '@safe/builder';
 import { GetFormByIdQueryResponse, GetPageByIdQueryResponse, GetStepByIdQueryResponse, GET_FORM_BY_ID, GET_PAGE_BY_ID, GET_STEP_BY_ID } from '../../../graphql/queries';
 import { Subscription } from 'rxjs';
+import {SafeSnackBarService} from '../../../../../../safe/src/lib/services/snackbar.service';
```
AntoineRelief: @MartinGbz ⚠️ not the correct way to import safe libraries

PR #303, opened 2021-09-27
AB#10781

MartinGbz:
# Description

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/projects/safe/src/lib/style/survey.scss
+++ b/projects/safe/src/lib/style/survey.scss
@@ -1,28 +1,25 @@
 ::ng-deep .sv-boolean:not(.sv-boolean--checked) {
-    .sv-boolean__switch {
-        background-color: rgba(64, 64, 64, 0.5) !important;
-    }
+  .sv-boolean__switch {
+    background-color: rgba(64, 64, 64, 0.5) !important;
+  }
 }
 ::ng-deep .sv_q {
-    padding-bottom: 0.5em !important;
+  padding-bottom: 0.5em !important;
 }
 
 ::ng-deep .sv_q_required_text {
-    color: red;
+  color: red;
 }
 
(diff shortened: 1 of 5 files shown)


projects/safe/src/lib/style/survey.scss
```diff
@@ -10,3 +10,19 @@
...
+}
+
+::ng-deep .sv_main .sv_q_erbox:not([style*="display: none"]):not([style*="display:none"]){
+  background-color: transparent;
+  border-width: 0;
+  font-weight: bold;
+  order: 2;
+}
+
+.sv_qstn .content-left {
```
AntoineRelief: @MartinGbz we should not apply the css to this specific question configuration

AntoineRelief:
@MartinGbz this looks almost good.
I think we're just missing how we determine the css of the question, to have its content flexible, and use the order property

I found this link: https://surveyjs.io/Examples/Library?id=survey-cssclasses#content-js

if we look at the code it seems possible to have a custom cssContent property, just need to determine how to set that

Capture d’écran 2021-09-27 à 16 57 19

projects/safe/src/lib/style/survey.scss
```diff
@@ -10,3 +10,19 @@
 ::ng-deep .sv_q_required_text {
     color: red;
 }
+
+::ng-deep .sv_main .sv_p_root .sv_row .sv_q.sv_qstn:first-child:last-child > .content-left {
```
AntoineRelief: @MartinGbz ⚠️ I will correct that, but as explained previously the .content-left class is too specific, and cannot be used. It is only available for some questions, if the admin sets that up

MartinGbz: because this method worked I forgot that .content-left was a problem

PR #378, opened 2021-11-03
AB#11693

Matthis-M-ReliefApps:
# Description

Until now, the way to prefill a form with a record was the following:
1. In grid settings, add an action button with "Go to next step" and "prefill form" option
2. Add the form you want to prefill after the grid in a workflow
3. Select the record(s) you want to prefill the form with
4. Click on the "go to next step" action button
Result => The user navigate to the next step of the workflow, the form, with prefilled fields

This PR aims to make the process much simpler and intuitive for the user, by changing the flow to the following:
1. In grid settings, add an action button with "prefill form" option, and select the form you want to prefill
2. Select the record(s) you want to prefill the form with
3. Click on the "prefill form" action button
Result => A model opens containing the form, prefilled with the selected record(s)

## Type of change

# How Has This Been Tested?

- [ ] In a grid, add an action button with aforementioned settings. Select a record, and use the button. The form modal should open with its fields prefilled
- [ ] In a grid, add an action button with aforementioned settings. Select multiple records, and use the button. The form modal should open with its fields prefilled and the possibility to fill multiple forms this way.

# Checklist:


projects/safe/src/lib/components/widgets/grid/grid.component.ts
```diff
@@ -1032,7 +1038,18 @@ export class SafeGridComponent implements OnInit, OnChanges, OnDestroy {
...
-        this.workflowService.storeRecords(records);
+
+
+        // Open a modal containing the prefilled form
+        this.dialog.open(SafeFormModalComponent, {
+          data: {
+            recordId: selectedRecords.map(x => x.id),
+            template: options.prefillTargetForm,
+            locale: 'en',
+            prefillRecords: records,
```
AntoineRelief, line 1052: @Matthis-M-ReliefApps I'd have built a similar logic, moving the logic of the nextStep to the SafeFormModal when passing the data.

nevertheless I don't think you need to pass recordId there, as we want to create a blank record, containing some data

projects/safe/src/lib/components/form/form.component.ts
```diff
@@ -103,43 +95,13 @@ export class SafeFormComponent implements OnInit, OnDestroy, AfterViewInit {
...
-          if (resourcesField && resourcesField.resource === records[0].form?.resource?.id) {
-            cachedData[resourcesField.name] = records.map(x => x.id);
-          } else {
-            this.snackBar.openSnackBar(NOTIFICATIONS.recordDoesNotMatch, { error: true });
-          }
-        }
-      });
-    }
-
-    if (this.form.uniqueRecord && this.form.uniqueRecord.data) {
```
AntoineRelief, line 129: @Matthis-M-ReliefApps not something you'd seen, but this is important to keep
In fact, there's a use case where we can tell that a form has an unique record per user / per role etc

AntoineRelief:
@Matthis-M-ReliefApps almost all good for this one, I only changed one thing:
in fact the prefill form should be one of the forms related to the current form ( so for ex, if current form refers to another form, then it's related form )
so instead of templates, we would use relatedForms

projects/safe/src/lib/components/form/form.component.ts
Matthis-M-ReliefApps: Ahah I was sure I had deleted at least one important thing while doing the "cleaning"

PR #391, opened 2021-11-09
AB#11439

Hugo-Plissonneau:
# Description

When creating a new record from a question resource, use the record selected to prefill the information.

## Type of change

# How Has This Been Tested?

- [x] I created a form with a question resource, and tested if my new record was prefilled.

# Checklist:


AntoineRelief:
@Hugo-Plissonneau I changed the prefill to be directly the data of the current form, instead of the id
this way, it leaves the possibility for new records to also prefill nested records, otherwise, as the id does not exist yet, it would lead to an error

PR #434, opened 2021-12-01
OORT-6

Maximo-reliefapps:
# Description

Header alignment done, loading now hides the page.
- Searchbars and action buttons aligned in all the pages.
- All the pages are hidden until the loading is done (Previously just some pages did this).

## Type of change


AntoineRelief:
@Maximo-reliefapps can you also check if it's possible to reduce the size of the other filters that appear on click ? ( the select - date range etc )

PR #439, opened 2021-12-01
OORT-11

Maximo-reliefapps:
Change the mat paginator to allow different page sizes
we can have : 10 / 25 / 50
it will change how many items we load from the back
must change:
applications page
forms page
resources page

api config page


projects/back-office/src/app/dashboard/pages/api-configurations/api-configurations.component.ts
```diff
@@ -89,16 +91,25 @@ export class ApiConfigurationsComponent implements OnInit, OnDestroy, AfterViewI
...
         variables: {
-          first: ITEMS_PER_PAGE,
+          first: neededSize,
           afterCursor: this.pageInfo.endCursor
         },
         updateQuery: (prev, { fetchMoreResult }) => {
           if (!fetchMoreResult) { return prev; }
           return Object.assign({}, prev, {
-            apiConfigurations: {
+            forms: {
```
AntoineRelief, line 112: @Maximo-reliefapps I will correct this one in the main branch, but it should not change 
I think you copied the code you had in forms page, but in fact it's breaking the pagination behavior to do that

PR #461, opened 2021-12-07
OORT-17

Maximo-reliefapps:
Badge component added, tested on storybook - Colors should be changed.


projects/safe/src/lib/components/ui/badge/badge.module.ts
```diff
@@ -0,0 +1,20 @@
...
+
+
+
+@NgModule({
+  declarations: [
+    SafeBadgeComponent
+  ],
+  imports: [
+    CommonModule,
+    MatChipsModule
```
AntoineRelief, line 14: @Maximo-reliefapps I think we can remove this module import

PR #502, opened 2021-12-20
OORT-32

Maximo-reliefapps:
**Use Kendo Angular Layout instead of Mat-layout for the widget grid**

CHANGES:
- Angular material grid layout has been changed to Kendo angular tile layout.
- Widgets headers have been removed since Kendo tiles already have a header.
- Removed resize option from floating-option component.
- Resize can now be achieved dragging the mouse on the right / bottom side of a tile.


AntoineRelief:
the bar on the left is over the header of the platform

Capture d’écran 2021-12-20 à 15 24 59

AntoineRelief:
we should keep the styling of the widgets. Right now the style is a bit different, it looks like the classic one of kendo

PR #510, opened 2021-12-23
Ab#16102

Hugo-Plissonneau:
# Description

Added a loop to remove all keys from the data object that won't be needed for the update.
It won't changes the value of a resource question when using the auto-modify rows from a nested object.

## Type of change

# How Has This Been Tested?

- [x] I reproduced the steps from the ticket after my changes and its now working well.

# Checklist:

### DIFF
--- a/projects/safe/src/lib/components/widgets/grid/grid.component.ts
+++ b/projects/safe/src/lib/components/widgets/grid/grid.component.ts
@@ -239,29 +239,21 @@ export class SafeGridWidgetComponent implements OnInit {
    */
   private promisedRowsModifications(modifications: any[], items: any[]): Promise<any>[] {
     const promises: Promise<any>[] = [];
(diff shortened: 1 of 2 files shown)


projects/safe/src/lib/components/widgets/grid/grid.component.ts
```diff
@@ -240,22 +240,20 @@ export class SafeGridWidgetComponent implements OnInit {
   private promisedRowsModifications(modifications: any[], items: any[]): Promise<any>[] {
     const promises: Promise<any>[] = [];
     for (const item of items) {
-      const data = Object.assign({}, item);
+      const update: any = {};
```
AntoineRelief, line 243: @Hugo-Plissonneau I simplified the logic there
instead of copying the data, we can just build the update, the result will be the same

PR #520, opened 2022-01-12
OORT-45

Maximo-reliefapps:
Ticket 45 solved
- Added a loading icon to edit-row modal
- Styled edit-row modal to be similar to view-modal

### DIFF
--- a/projects/safe/src/lib/components/form-modal/form-modal.component.scss
+++ b/projects/safe/src/lib/components/form-modal/form-modal.component.scss
@@ -1,16 +1,15 @@
 @import "../../style/survey.scss";
 
 :host {
-  display: block;
+  display: flex;
+  flex-direction: column;
+  height: 100%;
   position: relative;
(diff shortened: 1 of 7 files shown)


projects/safe/src/lib/components/form-modal/form-modal.component.scss
```diff
@@ -19,3 +19,28 @@
...
+  left: 0;
+  margin: auto;
+
+}
+
+.content {
+  width: 77vw;
+}
+
+.buttons {
```
AntoineRelief: @Maximo-reliefapps in most cases I'd recommend to avoid too generic names like that, as they can relate to a lot of different things in the application

AntoineRelief:
@Maximo-reliefapps about the modals, I edited the code you did, as the size can be controlled when creating the modal
putting same parameters everywhere the modal is called will solve the issue

PR #530, opened 2022-01-14
OORT-40

Maximo-reliefapps:
CHANGES:
 - Searchbar added to applications selection in front end
- Shown applications limited to 5

CHANGES 2:
- Search icon added
- No results text is shown on an unsuccessful search


AntoineRelief:
@Maximo-reliefapps a quick update about this one, can you add the magnifying glass icon in the input ?

you can also remove the mat-menu and try to build a component for that instead, and the mat-menu is very limited

AntoineRelief:
@Maximo-reliefapps a few updates about this ticket:
- I used the overlay library from angular, that enables us to create popover
- make sure to clean the code before putting in review, some parts of it were duplicated / unused
- try to reuse components ( in that case, button / input ) instead of rebuilding their styles

PR #532, opened 2022-01-17
OORT-50

AntoineRelief:
CHANGES:
 - A min height has been added to the widget grid (At a height of one row the content of the grid was completely hidden).
 - Fix the problem where the grid paginator was hidden.


AntoineRelief:
@Maximo-reliefapps instead of doing a single test on 'grid', I edited the code to have it per type of widget, so each widget can have a minimum size, and we would test it before confirming the change

PR #553, opened 2022-01-27
OORT-55

Maximo-reliefapps:
CHANGES:
 - Added a 'Last updated' text at the bottom of chart and map widgets.
 - Text displays the hour of the last time the widget has loaded the data.
 - Text is only displayed if data has been loaded and the mouse is hover the widget.


projects/safe/src/lib/components/widgets/chart/chart.component.html
```diff
@@ -15,4 +15,5 @@
     <safe-bar-chart #chartWrapper class="widget-content" *ngIf="settings.chart.type === 'bar'" [title]="settings.chart.title"
         [legend]="settings.chart.legend" [series]="series"></safe-bar-chart>
     <mat-spinner [diameter]="30" *ngIf="loading"></mat-spinner>
+    <p class="lastUpdate" *ngIf="lastUpdate && lastUpdate.length > 0">Last update: {{lastUpdate}}</p>
```
AntoineRelief: @Maximo-reliefapps please avoid this kind of class name
should be formatted in kebab-case, not camelCase

PR #555, opened 2022-01-28
OORT-67

Maximo-reliefapps:
CHANGES:
 - 'Create a resource' button added in resources tab
 - New create a resource window component created


AntoineRelief:
@Maximo-reliefapps 
when possible, add JSDOC comments to the class / methods to describe them
You can see the example in the resource modal form
JSDOC comments can be generated by the IDE, typing '/**' then enter

PR #565, opened 2022-02-09
AB#24740

Maximo-reliefapps:
CHANGES:
- Grid time inline edition now works (Format problem).
- Grid modal form edition fix.


AntoineRelief:
@Maximo-reliefapps you will see that instead of removing the survey?.clear call, I updated a file called widget
with the library we're using for surveys, we can build some code on top of it
some part of this code was erasing the time values on update

PR #570, opened 2022-02-14
OORT-92

Maximo-reliefapps:
CHANGES:
 - npm arcGIS packages were broking kendo style, changed with CDN import


AntoineRelief:
@Maximo-reliefapps from your changes, I noticed what was wrong
in the angular.json file, you can find additional configuration for the projects
the styles, is an array that will load scss before build
so, instead of loading leaflet style from npm package, I put that in the index.html files, loading from CDN
Your changes were working, but not a clean solution
this should be cleaner

PR #596, opened 2022-03-03
OORT-115

Maximo-reliefapps:
CHANGES:
 - Translation support for all notifications


AntoineRelief:
@Maximo-reliefapps ⚠️ the code of safe library was not building, due to a path reference to an file that would not exist anymore

PR #599, opened 2022-03-08
AB#24828

Maximo-reliefapps:
CHANGES (For aggregation builder):
 - Pipelines can be reordered with drag and drop.
 - The grid has been moved to the preview tab.
 - Minor style fixes.
 - JSON validation on custom text area.

ISSUES:
 - Preview grid logic moved to the service, but it's run from where it was before, not sure it's the best way to do it.
 - I have not been able to find a good way to refactor the aggregation calls, so I left them as they were.


projects/safe/src/lib/components/ui/aggregation-builder/aggregation-builder-forms.ts
```diff
@@ -167,3 +173,13 @@ export const createAggregationForm = (
       )
     ),
   });
+
+const jsonValidator = (control: AbstractControl): ValidationErrors | null => {
```
AntoineRelief, line 183: @Maximo-reliefapps tries to add JSDOC comment to explain the methods

projects/safe/src/lib/components/ui/aggregation-builder/pipeline/pipeline.component.html
```diff
@@ -1,8 +1,11 @@
-<mat-accordion>
-  <mat-expansion-panel *ngFor="let stageForm of pipelineForm.controls; let index = index ">
+<mat-accordion cdkDropList (cdkDropListDropped)="drop($event)">
+  <mat-expansion-panel *ngFor="let stageForm of pipelineForm.controls; let index = index " cdkDrag>
     <mat-expansion-panel-header>
       <mat-panel-title>
         {{'aggregation.stages.' + stageForm.value.type | translate}}
+        <mat-icon class="info"
```
AntoineRelief: @Maximo-reliefapps safe-icon should be preferred when possible

projects/safe/src/lib/services/preprocessor.service.ts
```diff
@@ -47,6 +47,15 @@ export class SafePreprocessorService {
       text = text.split('{today}').join(todayToString);
     }
 
+    // === NOW ===
+    if (text.includes('{now}')) {
```
AntoineRelief: @Maximo-reliefapps there's an error in the branch, be aware of what you base your branch on
this is not supposed to come in this release

PR #658, opened 2022-04-13
Skeleton loading for application's user page

LenaelleL:
# Description

The application's user page was stuck in "loading" mode. 
This happend only in version 1.3.0 but not in 1.2.0. By comparing the code, I realized that in the 1.3.0 the value of "loading" was initialized to true while that was the contrary in 1.2.0. That's why I changed that.
I also found a key translation issue in this page that I missed while doing my previous ticket.

## Type of change

# How Has This Been Tested?

- [x] Tried to load the page in Firefox and Chrome for different applications (my sandbox and also other apps). Works in both cases, we are not stuck anymore.

# Checklist:


AntoineRelief:
@LenaelleRelief I update the users.component to keep loading as true by default
the error was coming from the parent components, not providing the loading to the users.component

PR #666, opened 2022-04-20
Oort 186

LenaelleL:
# Description

Here, we added a fr.json file to allow for a french translation of Oort. It is mostly complete but some of the words are still in english because I couldn't find an accurate translation (pulljobs, workflow etc.).

## Type of change

# How Has This Been Tested?

- [x] Tested it by changing the default language to French, it worked no problem except for the fact that some of the text is coded directly in English and not translated. I'm thinking for example of the "items per page" text.

# Checklist:


alexisdelage:
## Translation choices

* No [object] -> *Aucun [object]* instead of *Pas d'[objet]*
* Channels -> *Canaux de diffusion*
* Clear -> *Annuler* (as proposed by @AntoineRelief)
* createdOn -> *Date de création*, instead of *Créé le* (the change is made to be gender neutral)
* template -> *modèle*
* Layout -> *Mise en page* (it seems to be the more explicit for the user in our context)
* Pull jobs -> *Tâches planifiées* (it is not the most exact, since *planned Tasks* is more general and global, but it is more understandable for the user and since we do not have other types of planned task in the application I think we can use it)
* Workflow -> we keep workflow since it is understandable for french people and we have no equivalent in french
* Custom -> *Personnaliser*
* group / unwind -> *Grouper* / *Dissocier* (words taken from the Microsoft Office applications)
* confirmation message -> *Voulez-vous vraiment supprimer...* (it is the general sentence used on computer os, for example windows)
* core -> *formulaire source*
* Inline edition -> *Modifier par ligne*

## `New [object]` modifications

Replace the syntax `common.newObject | translate : {name: common.<object>.one} | translate}` by the new syntax `models.<object>.new | translate` so as to translate "new" in masculine or feminine in french.

## Some issues noticed about translation

I have noticed some issues when I change the language in the interface :
* On the main *Back-Office* page, the titles of the sections in the left navigation menu are not updated when user change language via the language button
* When we open an application, the same language as in the previous page show up but the language selector button reset to the default language (english). The same issue exists when we go back to the main back-office panel

PR #673, opened 2022-04-25
AB#30214

Maximo-reliefapps:
# Description

Added kendo-editor to email action body and email preview body, with it you are able to edit how the email will look, text and table format capabilities, the table style is lost on its use. 

## Type of change

# How Has This Been Tested?

- [x] Visual testing
- [x] In code testing

# Checklist:


LenaelleL:
I changed the editor used for the email preview popup to make add more possibilities of style to the tables (color, borders, merging cells etc.). I used CKEditor 5 for Angular which is open-source. I used a custom version of the Classic Editor build that is located on the file 'ckeditor.js' in the safe/src folder. Unfortunately it does not pass the lint.

To make this work, I had to add "allowJs": true to the file tsconfig.json.

During the preview some of the style elements are contained directly into the HTML tags of the body so they are kept in the e-mail but some of them are contained in a general file specific to CKEditor which means they won't be kept in the e-mail. That is why we need to add these style elements directy into the tags. 

By default, if we don't modify anything, CKEditor will make the tables border-less. I changed that so that if we don't modify anything in the preview the tables will still have a black border of 1px. Nonetheless, if we modify something in the file but not the border style of the tables it won't have any borders. The default border weight is also heavier that it shows in the text editor. It's a choice I made that I could completely change if we want to keep the default 'no-border' style.

For now, I kept the Kendo text editor for editing the body of the text before the preview but I think it would be a good idea to change that for more harmony between the two editors. And maybe if there is other text editors in the platform, change that as well.

PR #686, opened 2022-04-28
Oort 213

alexisdelage:
# Description

Add date format in preferences: the user can choose in the preferences the date and time format he prefers, for example the english one (`MM/DD/YY`) or the french one (`DD/MM/YYYY`). The proposed formats are based on the languages defined in the environment file.

In the code, we only have to replace the `date` pipe in HTML files by the `safeDate` pipe to make the date following the user preference. The developer can then indicates a pre-defined format options ([defined here](https://angular.io/api/common/DatePipe#pre-defined-format-options)), and then Angular will render the date according to the pre-defined format indicated by the developer and the user preference.

Two new modules have been added to `safe`:
* `SafeDateModule`, which contains the `SafeDatePipe` to use the pipe in HTML files;
* `SafeDateTranslateService`, the service which will manage the user preference and save it in local storage

This pull request remove also the dict for obtaining the complete names for each languages, in the user preference modal, by using the `Intl.DisplayNames` class of JavaScript (see the [MDN documentation here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames)). Note that this method is only available since TypeScript `ES2020`, but we use `ES2018`, so it has an `any` type in the code to make it work. This method however still compatible with all modern browsers (except Internet Explorer, according to the MDN documentation).

Note about adding more languages: if want to add more languages, we have to register the corresponding locale in the `app.module.ts` files of the backend and frontend.

## Type of change

# How Has This Been Tested?

Change the date in User preferences, and see the date being updated (for example in the application list, and in dashboard widgets). 

# Checklist:


projects/safe/src/lib/components/preferences-modal/preferences-modal.component.ts
```diff
@@ -48,22 +36,82 @@ export class SafePreferencesModalComponent implements OnInit {
...
   }
+
+  /**
+   * Get the full name of a language from its code
+   *
+   * @param lang The code of the language we want the name of
+   * @param displayLang The code of the language in which we want the name translated
+   * @returns The language name
+   */
+  private getLanguageName(lang: string, displayLang: string): string {
```
AntoineRelief: @alexisdelage that's a good initiative, but I prefer leaving the name of a lang in its own language, as it's easier for an user to find it

PR #693, opened 2022-04-30
Oort-184

matheus-relief:
# Description

So initially I misunderstood the bug because of the title of the ticket. I thought that by "reload", you meant fetching the form again from the backend when it was saved (which was only happening in the app-preview, and that puzzled me for quite a while, I couldn't figure why that was happening. Eventually, I understood why, and fixed it (I wouldn't say it was such a waste of time, cause I learned quite a bit about apollo and graphql in the process). Than, I realized I that wasn't the bug at all, but I kept the change, cause it saves a request and we have all the updated data anyway. 
The wanted behavior, however, was achieved merely by removing an if statement that was doing exactly what we didn't want the form to do on submit.

## Type of change

# How Has This Been Tested?

- [ ]On submitting a form with configured as a unique record, the form no longer resets to its first available page 

# Checklist:

### DIFF
--- a/projects/back-office/src/app/app-preview/pages/form/form.component.ts
+++ b/projects/back-office/src/app/app-preview/pages/form/form.component.ts
@@ -84,13 +84,13 @@ export class FormComponent implements OnInit {
           .valueChanges.subscribe((res) => {
             this.page = res.data.page;
             this.apollo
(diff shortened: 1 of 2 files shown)


AntoineRelief:
@matheus-relief I reproduced your changes on 1.2

PR #707, opened 2022-05-10
AB#28021 (Part 1) Correct date formats in surveyjs

alexisdelage:
# Description

Correct the date formats in surveyjs date fields. 
To achieve this, we replace the default date fields (which are simply the default `<input type=date">` HTML element) by kendo datepicker fields in the `afterRender` function of `widgets.ts`. To make it work, we have to include many providers dependencies in `app.module.ts` (it does not work if it is only in the child module) and to include the `DateInputModule` in each module where a surveyjs form is displayed.
Concerning the date formats, the kendo datepicker work with the `Date` object, and we store in our database the date as sting in the ISO format. Then we give to the kendo datepicker which format to use for displaying the date to the user.

About the date in properties panel: to replace the "min" and "max" properties which also work with default date fields, we have to hide these properties and display two new properties which use the kendo datepicker and update on change the "min" and "max" hidden properties.

This PR is only the first part of the ticket: I will do another PR for the modifications about the date filters for grids.

NB: another PR has been made for the backend, so as to correct a bug on time fields.

## Type of change

# How Has This Been Tested?

Create a survey with a `date` field, and check if the kendo datepicker displayed correctly in different pages where the form is shown.

# Checklist:

# Screenshots

image
image
image


projects/safe/src/lib/style/survey.scss
```diff
@@ -24,6 +24,13 @@
   }
 }
 
+// === DATE / TIME INPUTS ===
```
AntoineRelief: @alexisdelage, when using ::ng-deep, you need to be aware that if you load it once, then it's applied everywhere in the platform

in general, we want to add some properties, that would tell that the ::ng-deep only applies for some cases ( nesting with classes etc )

alexisdelage: Ok, I will take care of it next time!

alexisdelage:
So I correct the issue with kendo pickers in surveys :

* **Description of the issue :** When we display a survey with kendo date/time/datetime pickers, the pickers are not rendered until we click somewhere on the page. This issue does not appear in the survey builder
* **Components affected :** `form.component`, `form-modal.component`, `record-modal.component`
* **Solution found :** Force to "refresh" the components by executing an empty method after the page is loaded, with `setTimeout(() => {}, 500);`. I tried to use the angular lifecycle hooks and surveyjs events instead of the `setTimeout` method, but none of them worked.
* **Is it a good solution?** No
* **Does it work?** Yes
* **Why does it work?** I don't know

PR #708, opened 2022-05-11
AB#28021 (Part 2) Correct date formats in grid filters

alexisdelage:
# Description

Correct the format of date values in filter settings for grids. We use the material date picker instead of a text input, so as to select the date format (in this case, I used the format `en-GB` which corresponds to `DD/MM/YYYY`).

> Some notes about time inputs: concerning the time inputs, I noticed some strange behavior regarding the timezone (it is not really a bug since it could be what is expected, but it is not very understandable for the user so maybe we can do something). When a user fill a form, the input datetime or time (a time is in fact a datetime with a day set on 01/01/1970) is considered in the user timezone (defined by the user browser), and we store it in our database with this timezone (in the ISO format). But when we display datetimes in a grid, the timezone is not taken into account, and the datetime is so displayed in GMT.
> 
> For instance, if a French user enter the time `15:17`, the timepicker in the survey convert it to a `Date` object with the user browser timezone, and so the ISO format of this time is `"1970-01-01T14:17:00.000Z"`. But when we show it in a grid, the user browser timezone is not taken into account and so we display the time `14:17`, even if it is the same user. Maybe we could fix it by simply indicating in the column header that the time is in GMT, so the user knows why it is not the same time (and it can be useful if the survey is answered from multiple timezones).

## Type of change

# How Has This Been Tested?

Add a filter on a date field of a survey. Check in preview (and then in the grid) that the filtered data are correct.

## Sreenshots

image

# Checklist:


projects/safe/src/lib/components/query-builder/tab-filter/tab-filter.component.html
```diff
@@ -152,6 +140,15 @@
...
               </ng-container>
+              <!-- matDatepicker input must be direct child from mat-form-field -->
+              <input
+                *ngIf="['Date', 'DateTime'].includes(selectedFields[i].type)"
+                matInput [matDatepicker]="mypicker" formControlName="value">
+              <mat-datepicker-toggle
+                *ngIf="['Date', 'DateTime'].includes(selectedFields[i].type)"
+                matSuffix [for]="mypicker">
+              </mat-datepicker-toggle>
+              <mat-datepicker #mypicker></mat-datepicker>
```
AntoineRelief: @alexisdelage just as a general comment for future tickets, I'm not a big fan of the 'mything' names in the code ;)

AntoineRelief:
Capture d’écran 2022-05-30 à 15 31 29

@alexisdelage this is what I was referring to

I added an expression capacity, so user can either pick the date, or build one ( and so, use methods like today() )

PR #716, opened 2022-05-16
AB#27480

matheus-relief:
# Description

This PR changes the frontend so it fetches the record history (all the versions from the recrod), parses it and displays it properly. It also adds an option for the user to filter the history by field.

## Type of change

# How Has This Been Tested?

Fetching the record history from the backend can be tested by opening the history option for a record in a grid widget.
Filtering by field is tested by selecting one of the fields in the field dropdown of the history component. 

## Sreenshots
image

# Checklist:

- [x I have commented my code, particularly in hard-to-understand areas


matheus-relief:
@AntoineRelief I updated the filter by field logic as you requested, but we have a problem. I didn't realized earlier, but I remember now one of the reasons I was originally using the structure instead of the fields of the form directly is because only in the structure I had access to the title of the questions (instead of only the names). Obviously the same problem is happening in the backend now too.

Besides the title, the rows and columns titles aren't being saved as expected for the 'dynamicmatrix' question type. (it is correct on the structure though, that's why it broke after the change)
Also, I realized we have a problem in the backend, as it is, we're not considering the user's timezone to get the display values, but I'm not sure the best way to go about it... send the timezone as a param?

PR #730, opened 2022-05-19
OORT-240 fix canUpdate and canDelete in toolbar for grid

LenaelleL:
# Description

There was an issue that was discovered while working for the Guyane prefecture that even though the user from a specific role had no authorization to update or delete the records, the buttons to convert and delete rows were still showing when selecting rows.

 It is now fixed by changing the amount of data that was passed between the grid and the toolbar components. Before, the toolbar component that controlled whether or not these buttons were shown did not have access to anything else than the IDs of the selected rows and thus was not able to tell if the user could delete/update or not.

There was also some code duplication between the core-grid and the grid component that was causing weird issues but the useless part is now deleted.

## Type of change

# How Has This Been Tested?

- [x] This has been tested by reproducing locally (directly into the DB) on a test survey the configuration we had on the Guyane survey and setting up a grid with the data from this survey. 

## Screenshots

image

# Checklist:


projects/safe/src/lib/components/ui/core-grid/toolbar/toolbar.component.ts
```diff
@@ -23,11 +23,11 @@ export class SafeGridToolbarComponent implements OnInit {
   }
 
   get canUpdate(): boolean {
-    return !this.items.some((x) => x.canUpdate);
+    return !this.items.some((x) => !x.canUpdate);
```
AntoineRelief, line 26: @LenaelleRelief I think this is the correct logic
we need to check if there is no record that is the user cannot update

so we first try to check if one record is not updatable ( .some )
then we return the opposite of this first result

so if one record like that exists -> false
then -> true

PR #742, opened 2022-05-23
OORT-251

Maximo-reliefapps:
# Description

Added translation support for tinymce editor.
The translation files are stored in the i18n folder.
A pair array has been added to email.ts const so the app languages are correctly assigned to its tinymce correspondent ones.

## Type of change

# How Has This Been Tested?

- [x] Has been tested visually in app.

## Screenshots
Screenshot_20220523_094430

# Checklist:


AntoineRelief:
@Maximo-reliefapps a few changes for this ticket:
- current language should come from the translation service ( it's the API we have internally ). At the end, it's kind of the same that trying to get the language from the local storage, but it is cleaner to do it with the service
- I removed all languages we don't use. It is not relevant to add files that we don't use

PR #750, opened 2022-05-25
Ab#29318

LenaelleL:
# Description

We have validators for forms that were not taken into account when doing inline edition in the grid widgets. Now we use custom SurveyJS validators to check the data taken from the form structure attribute. If there is an error, a red snack-bar describing the error is shown on the user's screen.

In the core-grid component, we create a query to load the form structure that contains the validator info into the metadata of the gridData.
Then we create a function that modifies that structure to make its format a easier to treat. 
We then use the custom SurveyJS validators to check the data for the different types of validators. They all work rather instinctively except for the expression validator that requires the data to be put into an object in order to check it. There is no documentation for these SurveyJS validators otherwise.

We also check that the cell is not empty if the field is mandatory.

## Type of change

# How Has This Been Tested?

- [x] Tested by using all 5 types of validators on a form and displaying the data from this form in a grid. We then modify this data and we can check if it is valid or not.

## Screenshots

image

# Checklist:


projects/back-office/src/app/app-preview/pages/workflow/workflow.component.ts
```diff
@@ -1,19 +1,20 @@
...
   GET_WORKFLOW_BY_ID,
 } from '../../../graphql/queries';
 import { PreviewService } from '../../../services/preview.service';
 import { TranslateService } from '@ngx-translate/core';
+import {
+  ContentType,
+  SafeSnackBarService,
+  Step,
+  Workflow,
+} from 'projects/safe/src/public-api';
```
AntoineRelief, line 17: @LenaelleRelief not sure where that's coming from, but this is something that we really don't want

all imports from the safe project, if not in safe project itself, should come from '@safe/builder', which can be considered as the entrypoint of the library

projects/safe/src/lib/components/ui/core-grid/grid/grid.component.ts
```diff
@@ -485,6 +490,8 @@ export class SafeGridComponent implements OnInit, AfterViewInit {
    * Saves edition.
    */
   public onSave(): void {
+    // eslint-disable-next-line no-underscore-dangle
+    const stringVal = this.data.data[0]._meta.formStructure;
```
AntoineRelief, line 494: @LenaelleRelief not a valid variable name, as it does not tell what the variable represents
we know it's a string value, but we don't know a value of what

projects/safe/src/lib/components/ui/core-grid/grid/grid.component.ts
```diff
@@ -534,6 +552,251 @@ export class SafeGridComponent implements OnInit, AfterViewInit {
...
+  checkValidation(validator: any): string {
+    let errMessage = '';
+    let row = 0;
+    const data = this.data.data;
+    Object.keys(data).forEach((value1, key1) => {
+      Object.keys(data[key1]).forEach((value2, key2) => {
+        Object.keys(validator).forEach((value3, key3) => {
+          if (value2 === validator[value3].valueName) {
+            //Check if the field is required and if the cell did not became empty
+            if (validator[value3].isRequired && data[key1][value2] === '') {
```
AntoineRelief, line 601: @LenaelleRelief that only works for string values

projects/safe/src/lib/components/ui/core-grid/grid/grid.component.ts
```diff
@@ -47,6 +47,9 @@ import { SafeDownloadService } from '../../../../services/download.service';
 import { debounceTime, distinctUntilChanged } from 'rxjs/operators';
 import { SafeExportComponent } from '../export/export.component';
 import { GridLayout } from '../models/grid-layout.model';
+import { SafeSnackBarService } from 'projects/safe/src/public-api';
```
AntoineRelief, line 50: @LenaelleRelief similar comment, we should not import like this in safe library
the import must use a relative path, if the code to import is from safe library itself
it should start with ./ or ../

AntoineRelief:
@LenaelleRelief partially duplicated in this [PR](https://github.com/ReliefApplications/oort-frontend/pull/772)

The new PR directly puts the data in a survey in order to validate it
The logic simplifies a lot the feature, as it does not require to create a block per type of validation
everything is directly handled by surveyjs

PR #752, opened 2022-05-26
Oort 271 - Custom popup for maps

LenaelleL:
# Description

Before the latitude and longitude fields could only be chosen from the selected fields. Here we allow them to be selected from all the fields so that the selected fields are only the ones that show up on pop-ups.

In the case of maps, we also rename the tab "Fields" to "Pop-up" and add a little explanation message so that the users understand that these selected fields are the ones that are shown in the pop-up.

I based this ticket on Maximo's OORT-195 ticket so that I could have the different new tabs and rename one pop-up.

## Type of change

# How Has This Been Tested?

- [x] Tested on a map widget with data that is correctly shown on markers popup

## Screenshots
image
image
image

# Checklist:

### DIFF
--- a/projects/safe/src/lib/components/widgets/map/map.component.ts
+++ b/projects/safe/src/lib/components/widgets/map/map.component.ts
@@ -127,7 +127,10 @@ export class SafeMapComponent implements AfterViewInit, OnDestroy {
     this.drawMap();
     // Gets the settings from the DB.
     if (this.settings.query) {
-      const builtQuery = this.queryBuilder.buildQuery(this.settings);
+      const builtQuery = this.queryBuilder.buildQuery(
+        this.settings,
+        'latitude\nlongitude'
+      );
       this.dataQuery = this.apollo.watchQuery<any>({
         query: builtQuery,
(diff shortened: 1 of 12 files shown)


projects/safe/src/lib/components/widgets/map/map.component.ts
```diff
@@ -127,7 +127,10 @@ export class SafeMapComponent implements AfterViewInit, OnDestroy {
     this.drawMap();
     // Gets the settings from the DB.
     if (this.settings.query) {
-      const builtQuery = this.queryBuilder.buildQuery(this.settings);
+      const builtQuery = this.queryBuilder.buildQuery(
+        this.settings,
+        'latitude\nlongitude'
```
AntoineRelief, line 132: @LenaelleRelief that cannot work because it is not needed that the form contains fields named as latitude / longitude, it's more flexible than that

LenaelleL: Sorry I think i misunderstood the ticket then...

PR #755, opened 2022-05-30
OORT-259 Create multiple series in charts

alexisdelage:
# Description

This PR allows to use multiple series in line, bar, and column charts.

It also renames `xAxis` and `yAxis` to  `category` and `value` to fit bar, pie and donut charts.

Add a new field `Series` in the chart settings for line, bar, and column charts.

Add a new field for the group stage on the pipeline aggregation to allow to create groups indexed on multiple fields (two in this case).

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

## Breaking change

The fact that `xAxis` and `yAxis` have been renamed introduce a breaking change: charts created with previous versions will keep the `xAxis` and `yAxis` keys instead of the new ones.
In practice, this won't affect the display of existing charts: there is an operation in the backend to take into account this old keys.
However, when we open the settings of a chart, the selected fields will not appear in the `Category` and `Value` fields at the bottom of the modal (this is the only bug). The user have to select them again for being able to save the changes.

# How Has This Been Tested?

To test series in charts:
* Create a new form, for example `Cars`, with fields `id`, `type`, and `year`
* Create some records (enough to have significant data to display)
* Create a new dashboard and add a widget of type `Bar`, `Column`, or `Line` (for our example `bar` or `column` will be more convenient)
* Select the form `Cars` and all the previous fields created on the form
* In the aggreagtion tab, create a "Group" stage
* Select the `year` field in the "Group by" section, click the **+** button, and select the `type` field
* Click the "Add field" button and create a new field named "count" for example, with the "Count" operator
* At the bottom of the modal, select `Year` as the Category, `Count` as the value, and `Type` as the Serie.
* Save your changes and you should see a beautiful chart with series!

## Sreenshots

image
image

# Checklist:

### DIFF
--- a/projects/safe/src/i18n/en.json
+++ b/projects/safe/src/i18n/en.json
@@ -249,9 +249,9 @@
 			"fieldName": "Field name",
 			"groupBy": "Group by",
 			"mapping": {
-				"series": "Series",
-				"xAxis": "X Axis",
-				"yAxis": "Y Axis"
+				"category": "Category",
+				"seriesItem": "Serie",
+				"value": "Value"
 			},
 			"operator": "Operator",
 			"operators": {
--- a/projects/safe/src/i18n/en.json
+++ b/projects/safe/src/i18n/en.json
@@ -249,6 +249,7 @@
 			"fieldName": "Field name",
 			"groupBy": "Group by",
 			"mapping": {
+				"series": "Series",
 				"xAxis": "X Axis",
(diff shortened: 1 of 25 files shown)


AntoineRelief:
@alexisdelage solved in:

https://github.com/ReliefApplications/oort-frontend/pull/788

I reproduced the logic with some modifications

PR #773, opened 2022-06-06
OORT-288

matheus-relief:
# Description

This PR adds a button to the `SafeDropdownFilterComponent` that clears the filter selection.

## Type of change

# How Has This Been Tested?

In a grid configured to a resource that has a dropdown/radio type question, toggle the filters and select a value for that question. Upon that, a button to clear the selection should appear. On clicking it, the filter selection will be cleared.

# Checklist:

### DIFF
--- a/projects/safe/src/lib/components/ui/core-grid/dropdown-filter/dropdown-filter.component.scss
+++ b/projects/safe/src/lib/components/ui/core-grid/dropdown-filter/dropdown-filter.component.scss
@@ -1,3 +1,3 @@
-::ng-deep .k-filtercell{
-  gap: 0.3em;
+:host {
+  gap: 4px;
 }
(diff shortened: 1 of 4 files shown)


projects/safe/src/lib/components/ui/core-grid/dropdown-filter/dropdown-filter.component.scss
```diff
@@ -0,0 +1,3 @@
+::ng-deep .k-filtercell{
```
AntoineRelief: @matheus-relief I prefer to use px instead of em, as we're mostly using px in the code
+ avoid to use ::ng-deep like that, because if you do so, that will affect ALL elements with class .k-filtercell
in that case, there is a simpler solution, using :host

PR #776, opened 2022-06-07
OORT-255 refactor widgets for suveyjs to correct tagbox widget issue

alexisdelage:
# Description

**The original issue:**
* We can't use our custom properties (like tooltip) on tagbox questions

**The reason:**
* Our custom properties are defined inside a widget, and tagboxes are defined inside another widget

**The original solution (previous PR):**
* Recreate the tagbox question inside our custom widget, by inheriting the question of the tagbox widget
* Problem: this is not compatible with other questions which depends of the tagbox widget, since we have to change the name of the tagbox type

**How we are supposed to use widgets:**
* according to the surveyjs documentation, we are supposed to define a widget for each custom question
* components are the same thing that widgets, just with an easier syntax and a little bit less functionalities (they don't allow to override an existing question, for example)

**What this PR does:**
* Split the current global widget into 2 widgets:
   - one widget for the text question (which manage the kendo editor for the date input type of the text question)
   - one widget for the comment question (to add the allowEdition button)
* Move the render methods of the `resource` and `resources` components into the `afterRender` method of each component (no need to define them in another place)
* Move the initialisation and rendering of the global custom properties (like tooltip) into global methods, which are called directly on the survey object so as to be applied on all questions (disregarding the widget)

**Some additional notes:**
* There was a confusion in the code about whether we use the `Survey` library or the `survey` instance in some places. I rename in the components and the widgets all the `survey` parameters into `Survey` when it was in fact the library (either the Survey Library or the SurveyCreator library).
* On contrary to widgets and components, the rendering of custom properties cannot be defined onto the `Survey` library object but only on the instance, so I had to add a call to the rendering method after each instance creation of a survey (that means in the form builder service and the form builder component).
* The parameter `element` of the `afterRender` method of components is not the same as the one for the widgets, so I added a querySelector in the rendering methods of `resource` and `resources` to make them work in their component

## Type of change

# How Has This Been Tested?

Create a form with a tagbox, add a tooltip and check it is displayed.
Also tested texts, resource and resources question to be sure that the re-factoring did not impact them. 

## Screenshots
image

# Checklist:


AntoineRelief:
to check during review:

- [x] resource / resources display not working after saving form
- [x] url not displayed correctly
- [ ] _multi text not displayed correctly ( not linked to PR )_
- [x] url display not working after saving form
- [x] tooltip not working after saving form

AntoineRelief:
to check during merge:

- [x] export resources option
- [x] ref data
- [x] action button in resource / resources is removed / replaced on translation change
- [x] translation
- [x] placeholder for resource
- [x] translation of tooltip
- [x] edit comment is removed / replaced on translation change

PR #792, opened 2022-06-13
AB#34839

Maximo-reliefapps:
# Description

Created the basic components/skeleton of the new widget  "summary card".
New components have been added to the project, `safe-summary-card`, `safe-summary-card-settings` and `safe-card-settings`.
For now there isn't any functionality, you are only able to add simple cards and change titles in the settings.

An icon and a color is missing, map color and icon have been used for now.

## Type of change

## Screenshots

Screenshot from 2022-06-13 09-38-07
Screenshot from 2022-06-13 09-38-25
Screenshot from 2022-06-13 09-38-52

# How Has This Been Tested?

- [x] All the changes have been tested manually in app

# Checklist:


projects/safe/src/public-api.ts
```diff
@@ -60,6 +60,9 @@ export * from './lib/components/widgets/map/public-api';
 export * from './lib/components/widgets/map-settings/public-api';
 export * from './lib/components/widgets/scheduler/public-api';
 export * from './lib/components/widgets/scheduler-settings/public-api';
+export * from './lib/components/widgets/summary-card/public-api';
```
AntoineRelief: @Maximo-reliefapps that's not needed
confusing because we put the other widgets in this file, but it's also not needed for the other widgets
only useful if we need to put the widgets in components outside of the shared library, but in fact we only use the widget in the safe-widget component, and safe widget component is only used in components in the library
at the end, we just need to expose the parent components that use the widgets, and that the other projects will use

projects/safe/src/lib/components/widgets/summary-card/summary-card.module.ts
```diff
@@ -0,0 +1,12 @@
+import { LayoutModule } from '@progress/kendo-angular-layout';
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+import { SafeSummaryCardComponent } from './summary-card.component';
+
+/** Module for scheduler  component */
```
AntoineRelief: @Maximo-reliefapps aware when duplicating code, the comments are not good

PR #846, opened 2022-07-04
OORT-387

alexisdelage:
# Description

Correct a bug in the tab filter when we remove a field from a form.
Correct the `createFilterGroup` method by removing an unused argument.
Correct another bug on fields removing: the fields which follow the removed one were not available in the resource anymore (see backend).

## Type of change

# How Has This Been Tested?

Try to create a form with some fields, then create a widget with a tab filter using one of the fields, and then delete that field in the form. Before, the filter tab then displays itself in a wrong way, now it displays an empty field to show that there was a field here before which is not available anymore (I preferred to show an empty field rather than deleting the field, because deleting the field implies to remove it from the structure object in the database and so to analyse and rebuild every filter structures on field deletion).

# Checklist:


AntoineRelief:
@alexisdelage I only added a few 'get' from lodash calls

PR #879, opened 2022-07-19
OORT-410 next

alexisdelage:
See #878 

Refactor info icons used on map widgets (not present on the 1.3.0 branch)


AntoineRelief:
@alexisdelage closed and duplicated changes on main branch directly, as there was too many updates, on previous branch

in previous branch, updated:
- class name to be 'cursor-pointer' instead of previous clickable class
- set size as 18 for most info icons
- add some scss mixin to add scss to icons in radio / slide toggle / checkbox components

PR #882, opened 2022-07-21
Oort 412

alexisdelage:
# Description

Correct several bugs with forms which use other languages than english:
* Correct error on new question saving when the default language is different than english, or when we have more than one language
* Correct error on choices display in grids when the question is translated
* Correct error on duplicated names for questions

What is done in this PR?
* Export types of custom questions and properties added to surveyJS so as to use them in other parts of the code
* For the name validation on form saving (where we create the question name in snake_case from the question title), operate directly on the survey instance instead of the JSON object extracted from the survey. This way, the translation issues are directly handled by the Survey class and we can keep our code cleaner (as a general rule, we must remind ourselves to parse the JSON structure  as a Survey instance every time we need to read it or modify it to avoid problems).
* Replace the `toSnakeCase` function by the `snakeCase` function of lodash, which handle more cases (like accents for french for example) and always return a snake_case string (the snake_case checking after the conversion hence become irrelevant)
* Correct the display of choices for select type questions in grids, when the survey has several languages, by using the default language of the survey
* Add a checking of duplicated names before updating the survey with the name values converted as snake case, because before, when the name already exists for another question, the `valueName` was set to the survey and so even if we change the title we cannot change the valueName and so the error still appears

## Type of change

# How Has This Been Tested?

Set the interface in another language than English, create a new form and create some questions and change their titles. Click the 'save' button: it should work!

For the grid issue, add a selectbase question (like dropdown) with multiple translations for each item: it should display the default language of the form defined in the Survey properties.

# Checklist:

### DIFF
--- a/projects/back-office/src/app/dashboard/pages/advanced-settings/advanced-settings.component.ts
+++ b/projects/back-office/src/app/dashboard/pages/advanced-settings/advanced-settings.component.ts
@@ -90,7 +90,6 @@ export class AdvancedSettingsComponent implements OnInit {
             { error: true }
           );
         } else {
-          console.log('FOo');
           const setting = res.data.setting;
           const isNotLocal =
             setting &&
--- a/projects/safe/src/i18n/en.json
+++ b/projects/safe/src/i18n/en.json
@@ -436,7 +436,6 @@
 		},
 		"formBuilder": {
 			"errors": {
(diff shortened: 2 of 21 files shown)


AntoineRelief:
@alexisdelage I reverted the changes about surveyLang in the grid, and edited the setup of meta fields instead

PR #907, opened 2022-08-16
AB#39529 Bug of card items teleporting

alexisdelage:
# Description
Fix a bug where the first card item teleport itself when clicked on the header.

Cause: nesting tile-layouts. The tile layout kendo component is not supposed to be nested inside another one, so it breaks because of some interns stuffs .

Solution: prevent kendo to change the css properties on a click action, by adding the same properties with their values before the click and the *important* tag in a class (not on the component itself since the javascript code of kendo changes the value of the component properties).

## Type of change

# How Has This Been Tested?

Click on the header of the first card. Also tested with a button in the header.

# Checklist:

- [x] Linting does not generate new warnings


AntoineRelief:
@alexisdelage solved it in the main branch, by reproducing the logic of the kendo layout to the summary card, while in display mode

PR #1106, opened 2023-02-01
start to integrate clusters

AntoineRelief:
# Description

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Sreenshots

# Checklist:

( * == Mandatory ) 


AntoineRelief:
@estelafs 
I started to display a test cluster layer
What you can do on this one:
- check the new util-test and ensure that the points are not outside of the bounds of the map ( which is the case currently, I think the longitude are incorrect ?? )
- create a new file in the map component, called cluster-test.ts
- in this file, add a new method: generateClusterLayer -> we would need to put in it all the code about creation of the cluster group I put in the map component ( starting with the line // Clusters )
- this method should return the clusterGroup, so this line can stay in the map component:
`this.map.addLayer(clusterGroup);`
- in addition, demonstrate the custom icons for the cluster, in the same method, for the clusters + the icons
- make sure number of features can still appear on top of the cluster icon
- you're free to use any svg you would need for the icons
- see if we can control the layer visibility from the layerControl of leaflet + set opacity etc as we have for the other layers

estelafs:
@AntoineRelief 
- I changed the cluster icon but also updated the Integrate Clusters file in the Drive with some examples of other ways of how it can be done

- I didn’t find a way to control the layers and options of the cluster layer using the `leaflet.control.layers.tree` plugin used for the other layers.  We could use just the `L.control.layers()` and it works, but it'll create a new layer control and we'll have something like this:

```
var overlayMaps = {
  cluster: clusterGroup,
};
L.control.layers(undefined, overlayMaps).addTo(this.map);
```

layercontrol

We have some other options as well: 

- the sub-plugin for the cluster plugin [Leaflet.MarkerCluster.LayerSupport](https://github.com/ghybs/Leaflet.MarkerCluster.LayerSupport)
- and the [Leaflet.FeatureGroup.SubGroup](https://github.com/ghybs/Leaflet.FeatureGroup.SubGroup)

PR #1107, opened 2023-02-01
AB#44894 - start heatmap

AntoineRelief:
# Description

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Sreenshots

# Checklist:

( * == Mandatory ) 


projects/safe/src/lib/components/widgets/map/heatmap-test.ts
```diff
@@ -0,0 +1,61 @@
...
+  //   0.5: '#C01627',
+  //   0.75: '#DB5D01',
+  //   1: '#FAE800',
+  // },
+  max: 1.0,
+  radius: 10,
+  blur: 15,
+};
+
+export const generateHeatMap = (
```
AntoineRelief, line 41: @matheus-relief with this setup, I only get one color, even if I have thousands of points, can you have a look ?

projects/safe/src/lib/components/widgets/map/heatmap-test.ts
```diff
@@ -0,0 +1,61 @@
...
+  //   onEachFeature: (feature: any, layer: any) => {
+  //     layer.bindPopup('super test');
+  //   },
+  // }).addTo(map);
+  // https://github.com/Leaflet/Leaflet.heat for options
+  const heatMapLayer = L.heatLayer(
+    heatArray,
+    merge(defaultHeatMapOptions, options)
+  );
+  heatMapLayer.on('click', (event: any) => {
```
AntoineRelief, line 113: @matheus-relief can we find out if we can listen to click events on the layer ?
if possible, try to find points near the click, and display popup

projects/safe/src/lib/components/widgets/map/heatmap-test.ts
```diff
@@ -0,0 +1,61 @@
...
+  geoJSON.features.forEach((feature) => {
+    if (feature.geometry.type === 'Point') {
+      heatArray.push([
+        feature.geometry.coordinates[1], // lat
+        feature.geometry.coordinates[0], // long
+        get(feature, 'properties.weight', 1), // weight -> should be in properties.weight of the feature
+      ]);
+    }
+  });
+  // L.geoJSON(geoJSON, {
```
AntoineRelief: @matheus-relief this is a try I started, to see if we could not still have popup for the points of the GeoJSON but that seems to consume too much resource

projects/safe/src/lib/components/widgets/map/heatmap-test.ts
matheus-relief: Well, I think it's because the `maxZoom` and `minOpacity` options aren't set. From my testing, I found that `minOpacity: 0.5` and `maxZoom: 18` work fine, but I guess this could change based on the use case. I think it should be configurable by the user. I changed that and the gradient a bit and got this:
```
{
  gradient: {
    0: '#08d1d1',
    0.25: '#08d169',
    0.5: '#deba07',
    0.75: '#de6707',
    1: '#de0715',
  },
  max: 1.0,
  radius: 10,
  blur: 15,
  minOpacity: 0.5,
  maxZoom: 18,
}
```
image

Oh and also, I think know what you and Estela were talking about in the daily, about the bounds/coordinates. In the geojson, the coordinates geometry array are in the [LNG, LAT] format and the leaflet takes in the other order

projects/safe/src/lib/components/widgets/map/heatmap-test.ts
matheus-relief: Tried a solution, that seems to be way speedier than rendering hidden markers, but there's a problem with the radius. I tried for a while but no luck. I explained better in the comments, in the image all the circles have the same radius (in km). But because of the projection distortion, they appear to be bigger the number of point in them are correct, though.

image

PR #1129, opened 2023-02-14
Feat/ab#56096 added download menu on a component

MwanPygmay:
# Description

Made the download map component. It implements a download button in the bottom left of the screen with a menu of options. On click, the menu is displayed; When clicked on again, the menu is hidden. Updated Kendo since ng build safe could not run because of a kendo-related error. Upgrading all kendo-related packages fixed the problem.

## Type of change

# How Has This Been Tested?

Add a map widget. Click the download button. A menu appears. Click "Selected layers". A dropdown select menu appears. Click download again. The menu disappears.

## Sreenshots

Screenshot from 2023-02-14 09-15-16

# Checklist:

( * == Mandatory ) 


projects/safe/src/lib/components/ui/map/map-download/map-download.component.html
```diff
@@ -0,0 +1,68 @@
+<div class="downloadMenu">
+  <link
```
AntoineRelief: @MwanPygmay don't do that, we should not load external styles like that

AntoineRelief:
@MwanPygmay 
for styling you must reuse similar elements than the ones already used by the tool:
- tailwind
- material icons

you should not add new classes if you don't need to, and you should not add any new library ( like font awesome ) if you don't need to

In addition, you should never put static text in the tool
It must work with translation, using i18n

AntoineRelief:
@MwanPygmay to check:

leaflet stop events does not seem to work on my side ( using chrome ). If I click on the control, a popup will appear below ( when trying to open the control + when closing the control )

use formbuilder ( reactive form ) instead of ngModel. That should also enable us to have default values in the form ( by default -> current view + all layers )

you can also disable the download button when expanded, if the form group you would create in the component is not valid

package.json
```diff
@@ -45,34 +45,34 @@
...
-    "@progress/kendo-angular-dialog": "^11.0.0",
-    "@progress/kendo-angular-dropdowns": "^11.0.0",
-    "@progress/kendo-angular-excel-export": "^11.0.0",
-    "@progress/kendo-angular-grid": "^11.0.0",
-    "@progress/kendo-angular-indicators": "^11.0.0",
-    "@progress/kendo-angular-inputs": "^11.0.0",
-    "@progress/kendo-angular-intl": "^11.0.0",
-    "@progress/kendo-angular-l10n": "^11.0.0",
-    "@progress/kendo-angular-label": "^11.0.0",
+    "@progress/kendo-angular-buttons": "^11.1.0",
```
AntoineRelief, line 48: @MwanPygmay avoid to do package updates in tickets not related to that

AntoineRelief: you can, but you should tell why the update is done

PR #1204, opened 2023-03-22
fix/AB#59603_some_charts_legends_are_undefined

ValentinRLL:
# Description

Labels were sometimes undefined or null on charts. A filter was made on the series and aggregation configuration to ensure that they were not undefined.
Some fields are empty when selected fields in aggregation settings have been deleted.
It is no longer possible to delete a field if it is used.
An alert message has been added to inform users that they must remove the used fields before they can be deleted.
A filter has been added to avoid displaying empty fields for old aggregation configurations.

## Ticket

https://dev.azure.com/WHOHQ/EMSSAFE/_boards/board/t/App%20Builder%20-%20Dashboards/Stories/?workitem=59603

## Type of change

# How Has This Been Tested?

The chart had a null label at the beginning and no longer has a null label after the fix.
Some fields are empty at the beginning and are no longer empty after the fix.

## Sreenshots

image
image
Screenshot 2023-03-23 at 11 07 16
Screenshot 2023-03-23 at 11 07 58
Screenshot 2023-03-24 at 11 59 23

# Checklist:

( * == Mandatory ) 


libs/safe/src/lib/components/ui/aggregation-builder/aggregation-builder.component.html
```diff
@@ -1,12 +1,17 @@
 <safe-spinner *ngIf="loading"></safe-spinner>
 <ng-container *ngIf="!loading">
+  <safe-alert class="delete-fields-alert" *ngIf="someFieldsUsed">
```
AntoineRelief: @IMBiXx we want to avoid that
a class should not exist for one specific element, above all if the style of the class is quite simple
for this specific case, you should use tailwind classes

libs/safe/src/lib/components/widgets/chart/chart.component.ts
```diff
@@ -226,7 +226,15 @@ export class SafeChartComponent
             const aggregationData = JSON.parse(
               JSON.stringify(data.recordsAggregation)
             );
-            if (get(this.settings, 'chart.mapping.series', null)) {
+            // Check if any of the elements has a series values that is null or undefined
+            const hasInvalidSeriesValue = aggregationData.some((x: any) =>
```
AntoineRelief: @IMBiXx I don't think this would work
if fact, you could have a case where a serie would be undefined ( like, to group 'other' values

PR #1216, opened 2023-03-27
Feat/ab#60139 db view current layer edition changes

unai-reliefapp:
# Description

Layer editor connection to the map view added.
Updated some interfaces in order to match same data structure for layer form and backend layer model
Update the map layers service in order to make API calls to get layers only when no layers are loaded or another one is added to the data base

## Ticket

[Link to ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/60139?src=WorkItemMention&src-action=artifact_link)

## Type of change

# How Has This Been Tested?

Please refer to the screenshots below

## Sreenshots

Layer editor:
image
(Only one API call done to get all layers)

# Checklist:

( * == Mandatory ) 


libs/safe/src/lib/components/ui/map/layer.ts
```diff
@@ -223,63 +235,60 @@ export class Layer {
...
    * @param restService rest service to get the geojson
-   * @param apollo Apollo
+   * @param mapLayersService MapLayersService
    * @returns Observable of LayerSettingsI
    */
   public static async createLayerFrom(
     layerIds: string[],
     restService: SafeRestService,
-    apollo: Apollo
+    mapLayersService: SafeMapLayersService
```
AntoineRelief: @unai-reliefapp I'm not sure about this method
In fact, I have the feeling that it adds lot of complexity to the file

perhaps you should put it in the map layers service instead

because, if we want to reuse it in other places (like in the map-layer, see other comments), then we need to declare all services that would be used by this specific method

libs/safe/src/lib/components/ui/map/layer.ts
```diff
@@ -223,63 +235,60 @@ export class Layer {
...
+            } as MergedLayerInfo)
+        ),
+        // Get them into an array after all pipes are done
+        toArray(),
+        // And set those layers into the children of our hardcoded layer group
+        // @TODO it would be mapped later onto it's current layer type
+        map((mergedLayerInfo: MergedLayerInfo[]) => {
+          return {
+            name: 'group layers',
+            type: 'group',
```
AntoineRelief: @unai-reliefapp if possible the group should not be called 'group layers', 
it should be possible to create a group without any name ( this is what would be done by default )
Also, I'm not sure if it's the best approach to do a query that would load all layers, and then filter it to find layers we should actually load
i prefer a solution where we would do a query to get the layer from its id

libs/safe/src/lib/components/widgets/map-settings/map-layer/map-layer.component.ts
```diff
@@ -72,16 +75,64 @@ export class MapLayerComponent
...
+      this.setUpLayer();
+    }
     this.setUpEditLayerListeners();
   }
 
+  /**
+   * Set default layer for editor
+   */
+  private setUpLayer() {
+    this.currentLayer = L.geoJSON({
```
AntoineRelief: @unai-reliefapp see other comments related to how we load layers
there, we should be able to load the layer as it appears in the previous map

libs/safe/src/lib/components/ui/map/layer.ts
unai-reliefapp: Hi @AntoineRelief , if the map contains more than one layer it would be more than one query, and with this approach we already have the layers loaded for the  layer editor as well, that's why i was commenting about the bulk API to load layers passing multiple ids. If this case is fine for you i'll replace it with a single query for each id

EDIT: Also i'll add the cache feature to it

libs/safe/src/lib/components/widgets/map-settings/map-layer/map-layer.component.ts
unai-reliefapp: @AntoineRelief when adding a new layer that is not saved yet how it should be built? From where do I take the spatial references?

KR, Unai

libs/safe/src/lib/components/ui/map/layer.ts
unai-reliefapp: Hi @AntoineRelief , I agree. I was focused on creating it within the layer class but I think we can move this out to the map layers service. I tried to initialize those services instance in the method itself but i could not make it work.

PR #1258, opened 2023-04-12
bugfix/AB#35551_SA-HQ-and-all-RO-R4.29-additional-date-field-for-subs…

unai-reliefapp:
…tantiation-date fix: apply correctly visibleIf condition to surveyjs questions by checking accesibility only if the condition does not exist fix: create new survey from scratch(parse of empty string) refactor: delete unnecessary hidden condition, cast to !!(Boolean) already transforms undefined values to false

# Description

Updated the visibility of question in surveys only if visibleIf property is not present.
Also fixed the error when creating a new survey from scratch(parse an empty string throws an error)

## Ticket

[Link to ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/35551)

## Type of change

# How Has This Been Tested?

Create any form with questions containing conditional visibility based on other question values and create a record from that survey

## Sreenshots

# Checklist:

( * == Mandatory ) 

### DIFF
--- a/projects/safe/src/lib/services/form-builder/form-builder.service.ts
+++ b/projects/safe/src/lib/services/form-builder/form-builder.service.ts
@@ -88,12 +88,14 @@ export class SafeFormBuilderService {
       for (const f of fields.filter((x) => !x.automated)) {
         const accessible = !!f.canSee;
         const editable = !!f.canUpdate;
-        const hidden: boolean = (f.canSee !== undefined && !f.canSee) || false;
         const disabled: boolean =
           (f.canUpdate !== undefined && !f.canUpdate) || false;
         const question = survey.getQuestionByName(f.name);
         if (question) {
-          question.visible = !hidden && accessible;
+          //If no visibleIf question preset then we update the visibility for it
+          if (!question.visibleIf) {
+            question.visible = accessible;
+          }
           question.readOnly = disabled || !editable;
         }
       }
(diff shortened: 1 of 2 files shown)


projects/safe/src/lib/services/form-builder/form-builder.service.ts
```diff
@@ -88,12 +88,14 @@ export class SafeFormBuilderService {
...
         const accessible = !!f.canSee;
         const editable = !!f.canUpdate;
-        const hidden: boolean = (f.canSee !== undefined && !f.canSee) || false;
         const disabled: boolean =
           (f.canUpdate !== undefined && !f.canUpdate) || false;
         const question = survey.getQuestionByName(f.name);
         if (question) {
-          question.visible = !hidden && accessible;
+          //If no visibleIf question preset then we update the visibility for it
+          if (!question.visibleIf) {
```
AntoineRelief: @unai-reliefapp correct me if that's wrong. My understanding of the code is that the "accessible" is never taken into account for the field I think
It should be, as another layer of visibility. If we considerer that you don't have access to the field, then you should never see it, whatever the visibleIf condition !

unai-reliefapp: Hi @AntoineRelief , if that is the case, we have two conflicting conditions to display the question, even if we take care of the accessible property, whenever the visibleIf condition is triggered, the question will show up.

As a quick fix, could it be enough to delete the question from the survey instance if the user does not have access to it? We keep the visibility tighten up to the visibleIf of the survey itself and the user would not access the question if he/she cannot do that

But this can conflict if the not accessible question is required for example(dont know i an admin can do that, but still), making the survey uncompletable.

Best approach would be to integrate the visibleIf with this metadata in the form creator/editor itself, but it would take more time

**EDIT**: Looks like deleting a required question updates the validity of the survey, so is not a blocker, we can try this approach i think

AntoineRelief: okay for this approach then

PR #1270, opened 2023-04-18
Feat/ab#61944 abc implement breadcrumbs component

Bricecrx:
# Description

The feature added is a new breadcrump component linked the new ui library designed thanks to tailwind. The breadcrumps component has three inputs : display, seperator and breadcrumps (list of breadcrump, using same interface as in the safe library).

## Ticket

I can't paste the azure ticket link because I do not have access to Azure. Here is the ticket id however : [AB#61944](https://dev.azure.com/WHOHQ/2888b816-2d79-49ef-a86e-561e6cd456a0/_workitems/edit/61944)

## Type of change

- [X ] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

I used storybook to test the visual of the component created and the impact of the parameters used

## Sreenshots

# Checklist:

( * == Mandatory ) 

- [x] I have made corresponding changes to the documentation ( if required )

### DIFF
--- a/libs/ui/src/index.ts
+++ b/libs/ui/src/index.ts
@@ -7,3 +7,4 @@ export * from './lib/checkbox/checkbox.module';
 export * from './lib/flyout-menu/flyout-menu.module';
 export * from './lib/icon/icon.module';
 export * from './lib/select/select.module';
+export * from './lib/breadcrumbs/breadcrumbs.module';
--- a/libs/ui/src/index.ts
+++ b/libs/ui/src/index.ts
@@ -1,3 +1,5 @@
+import { from } from 'rxjs';
+
 export * from './lib/ui.module';
 
 // === TAILWIND COMPONENTS === //
@@ -8,3 +10,10 @@ export * from './lib/flyout-menu/flyout-menu.module';
 export * from './lib/icon/icon.module';
 export * from './lib/select/select.module';
 export * from './lib/breadcrumbs/breadcrumbs.module';
+
+// === ENUMS === //
+export * from './lib/breadcrumbs/enums/breadcrumb-display.enum';
+export * from './lib/breadcrumbs/enums/breadcrumb-separator.enum';
+
+// === INTERFACES === //
+export * from './lib/breadcrumbs/interfaces/breadcrumb.interface';
--- a/libs/ui/src/index.ts
+++ b/libs/ui/src/index.ts
@@ -1,5 +1,3 @@
-import { from } from 'rxjs';
-
 export * from './lib/ui.module';
 
 // === TAILWIND COMPONENTS === //
--- a/libs/ui/src/lib/breadcrumbs/breadcrumbs.component.html
+++ b/libs/ui/src/lib/breadcrumbs/breadcrumbs.component.html
@@ -1,0 +1,1 @@
+<p>breadcrumbs works!</p>
--- a/libs/ui/src/lib/breadcrumbs/breadcrumbs.component.html
+++ b/libs/ui/src/lib/breadcrumbs/breadcrumbs.component.html
@@ -1,9 +1,65 @@
+<ng-template #defineSep>
+  <!-- Check the display style and adapt accordingly -->
+    <div [ngSwitch]="display">
+      <div *ngSwitchCase="BreadcrumbsTypes.CONTAINED">
+          <svg
+            class="h-full w-6 flex-shrink-0 text-gray-200"
+            viewBox="0 0 24 44"
+            preserveAspectRatio="none"
+            fill="currentColor"
+            aria-hidden="true"
+          >
+            <path d="M.293 0l22 22-22 22h1.414l22-22-22-22H.293z" />
+          </svg>
+      </div>
+      <div *ngSwitchCase="BreadcrumbsTypes.FULL">
+          <svg
+            class="h-full w-6 flex-shrink-0 text-gray-200"
+            viewBox="0 0 24 44"
+            preserveAspectRatio="none"
+            fill="currentColor"
+            aria-hidden="true"
+          >
+            <path d="M.293 0l22 22-22 22h1.414l22-22-22-22H.293z" />
+          </svg>
+      </div>
+      <div *ngSwitchCase="BreadcrumbsTypes.SIMPLE">
+        <!-- Check the separator style and adapt accordingly -->
+        <div [ngSwitch]="separator">
+          <div *ngSwitchCase="BreadcrumbsSeparators.CHEVRON">
+              <svg
+                class="h-5 w-5 flex-shrink-0 text-gray-400"
+                viewBox="0 0 20 20"
+                fill="currentColor"
+                aria-hidden="true"
+              >
+                <path
+                  fill-rule="evenodd"
+                  d="M7.21 14.77a.75.75 0 01.02-1.06L11.168 10 7.23 6.29a.75.75 0 111.04-1.08l4.5 4.25a.75.75 0 010 1.08l-4.5 4.25a.75.75 0 01-1.06-.02z"
+                  clip-rule="evenodd"
+                />
+              </svg>
+          </div>
+          <div *ngSwitchCase="BreadcrumbsSeparators.SLASH">
+              <svg
+                class="h-5 w-5 flex-shrink-0 text-gray-300"
+                fill="currentColor"
+                viewBox="0 0 20 20"
+                aria-hidden="true"
+              >
+                <path d="M5.555 17.776l8-16 .894.448-8 16-.894-.448z" />
+              </svg>
+          </div>
+        </div>
+      </div>
+    </div>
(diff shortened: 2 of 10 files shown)


libs/ui/src/lib/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -0,0 +1,27 @@
...
+})
+export class BreadcrumbsComponent {
+  @Input() path: string[] = [];
+  @Input() seperator: BreadcrumbSeperator = BreadcrumbSeperator.CHEVRON;
+  @Input() display: BreadcrumbDisplay = BreadcrumbDisplay.SIMPLE;
+
+  BreadcrumbsTypes = BreadcrumbDisplay;
+  BreadcrumbsSeperators = BreadcrumbSeperator;
+}
+export enum BreadcrumbDisplay {
```
unai-reliefapp: Please move the enums or interfaces to a separate folder named enums or interface in the breadcrumbs folder, depends the case and export them in the `index.ts` file as explained in the `README` so we can access those outside the lib if needed without the need of importing the whole component

libs/ui/src/lib/breadcrumbs/breadcrumbs.component.stories.ts
```diff
@@ -0,0 +1,27 @@
+import { moduleMetadata, Story, Meta } from '@storybook/angular';
```
unai-reliefapp: As we are creating a new component, let's try to keep a correct syntax for the naming, please correct the sep**e**rator to sep**a**rator

libs/ui/src/lib/breadcrumbs/breadcrumbs.component.html
```diff
@@ -0,0 +1,98 @@
...
+  aria-label="Breadcrumb"
+>
+  <ol
+    role="list"
+    class="mx-auto flex w-full max-w-screen-xl space-x-4 px-4 sm:px-6 lg:px-8"
+  >
+    <li *ngFor="let pth of path" class="flex">
+      <div class="flex items-center">
+        <svg
+          *ngIf="!(path.indexOf(pth, 0) === 0)"
```
unai-reliefapp: More easy way to do it without so much logic involved is to use the Angular built in logic to see if the element is last or first: https://angular.io/api/common/NgForOf#local-variables

For example
`<li *ngFor="pth of path; let last as last" class="flex>`

And you can place the svg after the `<a>` element in all cases and show it only if `!last`

libs/ui/src/lib/breadcrumbs/breadcrumbs.component.html
```diff
@@ -0,0 +1,98 @@
+<nav
```
unai-reliefapp, line 26: Please give it a second thought and try to not duplicate so much html code, workaround with different classes but try to keep the html structure as simple as possible.

For example to use one separator or another you can use the ng-template element and you dont have to worry where to place those:
https://www.tektutorialshub.com/angular/ng-template-in-angular/

KR, Unai

libs/ui/src/lib/breadcrumbs/breadcrumbs.component.stories.ts
```diff
@@ -0,0 +1,40 @@
...
+    }),
+  ],
+} as Meta<BreadcrumbsComponent>;
+
+const Template: Story<BreadcrumbsComponent> = (args: BreadcrumbsComponent) => ({
+  props: args,
+});
+
+export const Primary = Template.bind({});
+Primary.args = {
```
unai-reliefapp: Please check how cases are provided in the stories:
https://github.com/ReliefApplications/oort-frontend/pull/1269/files#diff-fe72179fde370ba2f9cff113c38a25b3ac8b383f0d2c731719effa2f088b089b
And add one case per type of breadcrumb case

Thanks

unai-reliefapp (changes requested):

libs/ui/src/lib/breadcrumbs/breadcrumbs.component.ts
unai-reliefapp: Don't forget to export those in the `index.ts` file to use them outside the UI library!

libs/ui/src/lib/breadcrumbs/breadcrumbs.component.stories.ts
unai-reliefapp, line 1: Why `CONTAINED` and `FULL` display modes cannot use `SLASH` separator?
Out of curiosity, i didnt find any guide about that.

KR, Unai

libs/ui/src/lib/breadcrumbs/breadcrumbs.module.ts
```diff
@@ -0,0 +1,9 @@
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+import { BreadcrumbsComponent } from './breadcrumbs.component';
+
+@NgModule({
+  declarations: [BreadcrumbsComponent],
+  imports: [CommonModule],
```
unai-reliefapp, line 10: In order to export anything from the NgModule you have to add it to the `exports` property as an array 😸 In this case you'll have to add:
`exports: [BreadcrumbsComponent]` below the imports

Info: https://angular.io/api/core/NgModule#exports

libs/ui/src/lib/breadcrumbs/breadcrumbs.component.stories.ts
Bricecrx: Because it uses icon from certain shapes and there is only chevron icons. 
Moreover, it is as defined in the ticket

unai-reliefapp: True @Bricecrx , it only applies to the simple, but is not specified which is the default separator for the other cases, the chevron or the slash.

Address the other issues and we'll wait until @AntoineRelief feedback

Thanks,
KR, Unai

libs/ui/src/index.ts
```diff
@@ -1,3 +1,5 @@
+import { from } from 'rxjs';
```
unai-reliefapp: We have an infiltrated here
👀 
this import should be gone

libs/ui/src/lib/breadcrumbs/breadcrumbs.component.html
```diff
@@ -0,0 +1,89 @@
...
+    }"
+  >
+    <!-- For each breadcrumb in the breadcrumbs list -->
+    <li *ngFor="let breadcrumb of breadcrumbs; let f = first" class="flex">
+      <div class="flex items-center">
+        <!-- Check the display style and adapt accordingly -->
+        <div [ngSwitch]="display" class="flex">
+          <div *ngSwitchCase="BreadcrumbsTypes.CONTAINED">
+            <svg
+              *ngIf="!f"
```
unai-reliefapp: Much much better @Bricecrx , thanks
As we arrived so far, let's improve a couple of more things here to create a stunning breadcrum html 😉 

- The first ngClass in here:
Applies `flex` class for all types, so we can delete all conditions instead of the one for `FULL`, as this:

```
[ngClass]="{
   'border-b border-gray-200 bg-white': display === BreadcrumbsTypes.FULL
  }"
```
Would you agree?

- ngSwitch is a good solve for the different template rendering, but you see that the ngSwitch is inside the ngFor, so the ngSwitch would check what to render for each rendered breadcrumb item, which is a no needed performance loss(breadcrumb is never going to be very long but still). Please try to render the needed template using ngSwitch as you did but do the check outside the ngFor template and render the resolved template inside the ngFor template instead.

- The `*ngIf="!f"` is placed in each one of the template types, which means that no matter what template type we display, that condition should be checked. ¿Wouldn't be better to do that check a level up and do it just once and write it once instead of doing it in each of the template types?

- Finally, even if we are not checking the separator type for the other types(`FULL` and `CONTAINED`), try not to repeat the default separators for them and reuse the one that you already create for the chevron!!

Thanks, good luck!

unai-reliefapp (changes requested):

unai-reliefapp (approved):

unai-reliefapp:
@AntoineRelief  This one looks good to me 👍
I'll keep all without merging them until you give a double check for the first ones
KR, Unai

libs/ui/src/lib/breadcrumbs/enums/breadcrumb-display.enum.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Enumeration of possible display styles
+ */
+export enum BreadcrumbDisplay {
```
AntoineRelief, line 4: @unai-reliefapp @Bricecrx be aware with the enums to always use string values, otherwise, the value is the index in the enum by default
which would not make any sense when using the component in another component, as it would look like that:
<ui-breadcrumbs [display]="0"></ui-breadcrumbs>

PR #1330, opened 2023-05-10
GIS map: set datasource as required

estelafs:
# Description
If the layer datasource is not valid, we don't display the other tabs.

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Sreenshots
DATASOURCE

# Checklist:

( * == Mandatory ) 


libs/safe/src/lib/components/widgets/map-settings/map-layer/map-layer.component.ts
```diff
@@ -84,6 +84,30 @@ export class MapLayerComponent
     };
   }
 
+  /**
+   * Get the datasource valid status, so we only display other tabs if valid
+   *
+   * @returns boolean if we have all necessary data to proceed
+   */
+  public get datasourceValid(): boolean {
```
AntoineRelief, line 93: @estelafs no need to change it, I'll do, but you can use a getter there

AntoineRelief: or even better, calculate it on form changes, so we don't need to get the values on every refresh of the template ( which would be almost any second )

PR #1359, opened 2023-05-17
Refactor/ab#64116_abc-replace-table-component

Bricecrx:
# Description

With this ticket, all mat-table directive and associated ones calls by the application is replaced with uiTableWrapper directive call associated with cdkTable mecanism. MatTableModule calls are also deleted and replaced every time it is needed. 

Moreover few improvements have been made to the directive : 
- Styling is now made in the style.scss and use cdk calls to get the good components to style (there was issue with styling in renderer when table content was modified
- Horizontal scroll is now authorized when needed (when there is too many columns for example)

Last but not least, there was issues with the sort functionnality in two components (their structure was different from others that used sorting) : api-configuration and reference-data. They have been reworked to be able to use sorting.

## Ticket

I can't paste the azure ticket link because I do not have access to Azure. Here is the ticket id however : [AB#64116](https://dev.azure.com/WHOHQ/2888b816-2d79-49ef-a86e-561e6cd456a0/_workitems/edit/64116)

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

It has been tested through navigation within the application and verification of replacements

## Screenshots

image
image
image
image

# Checklist:

( * == Mandatory ) 


libs/safe/src/lib/components/notifications/notifications.component.html
```diff
@@ -16,23 +16,23 @@
 <div class="overflow-x-hidden shadow-2lg">
   <!-- Table scroll container -->
   <div class="overflow-x-auto">
-    <table *ngIf="!loading" mat-table [dataSource]="notifications" matSort>
```
unai-reliefapp, line 19: Hi Brice, double check this `matSort` deleted because even if the directive is not watched from the .ts file, it's used to set default sorting behavior to the headers with the `mat-sort-header`

Files were I found this issue:
`distribution-lists.component`
`mapping.component`
`notifications.component`
`group-list.component`
`templates.component`
`user-list.component`

But could be more. My recommendation: create a `uiSort` directive that would receive the whole data array as input and save it in a private property, and this directive would listen to the sort changes from the `uiTableWrapper` and applied the needed sorting logic and set that new sorted array to the `[dataSource]` of the cdkTable and re render the rows

With this you can forget about any queries or especialized functions for each file. Make it work with plain data(not nested values, e.g.  `createdAt.month.day` and let's see that after you finish this first version to handle nested properties

unai-reliefapp: Double checked with @AntoineRelief . this deleted directives from the components are correct, you can forget about this comment @Bricecrx

libs/ui/src/style/styles.scss
```diff
@@ -314,6 +314,23 @@ body {
       @apply transition-transform;
     }
   }
+
+  .cdk-cell {
```
AntoineRelief: @unai-reliefapp @Bricecrx 
not a big fan of the scss there + the name of directives
wouldn't it be possible to extend the cdk table directives, like:
uiCell / uiHeaderCell
and then to apply the classes as we do in other places, with the directive?

unai-reliefapp: Hi @AntoineRelief , it was changed to this because whenever a table was sorted and new elements were added to the table after it's initial lifecycle hooks the classes were not applied to the new rows, and to avoid more logic in the directive and take advantage of the scss I ask him to do it like this.

If we add a more precise selector for a bigger specifity in the scss that could be enough or you rather go for the directive classes?

PR #1400, opened 2023-05-30
Refactor/oort 628 untyped form to typed form

ghost:
# Description

Changed all UntypedForm to TypedForm except 2 cases:
 - safe/src/lib/components/ui/core-grid/expanded-comment
 - safe/src/lib/components/graphql-select
where the form needed to be nullable.

In those 2 instances, I casted null to undefined where needed and undefined to false

## Ticket

[Ticket 628](https://oortcloud.atlassian.net/jira/software/projects/OORT/boards/3?selectedIssue=OORT-628)

## Type of change

# How Has This Been Tested?

- [ ] I tried to navigate to every page and it worked fine

# Checklist:

( * == Mandatory ) 

### DIFF
--- a/apps/back-office/src/app/application/pages/add-page/add-page.component.ts
+++ b/apps/back-office/src/app/application/pages/add-page/add-page.component.ts
@@ -1,10 +1,6 @@
 import { Apollo, QueryRef } from 'apollo-angular';
 import { Component, OnInit, ViewChild } from '@angular/core';
-import {
-  UntypedFormBuilder,
-  UntypedFormGroup,
-  Validators,
-} from '@angular/forms';
+import { FormBuilder, FormGroup, Validators } from '@angular/forms';
 import { MatLegacyDialog as MatDialog } from '@angular/material/legacy-dialog';
 import {
   ContentType,
@@ -58,7 +54,7 @@ export class AddPageComponent
   @ViewChild('formSelect') formSelect?: MatSelect;
 
   // === REACTIVE FORM ===
-  public pageForm: UntypedFormGroup = new UntypedFormGroup({});
+  public pageForm: FormGroup = new FormGroup({});
   public step = 1;
 
   /**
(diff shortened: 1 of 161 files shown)


apps/back-office/src/app/application/pages/add-page/add-page.component.ts
```diff
@@ -58,7 +54,7 @@ export class AddPageComponent
   @ViewChild('formSelect') formSelect?: MatSelect;
 
   // === REACTIVE FORM ===
-  public pageForm: UntypedFormGroup = new UntypedFormGroup({});
+  public pageForm: FormGroup = new FormGroup({});
```
AntoineRelief: @JulienKerlero 
in all places like that:
= new FormGroup({});
where we tell that the form is by default an empty one, you can check, when possible, by that:

pageForm!: FormGroup;

that will work if the formgroup is initialized in the ngOnInit of the component

matheus-relief (commented):
Although this is a good idea, it doesn't help us that much, as in most cases we still get no useful typings... I'll use the first changed file: `add-page/add-page.component.ts` to illustrate how this could be improved.

Instead of doing this
```ts
// === REACTIVE FORM ===
public pageForm!: FormGroup; // This is the same as FormGroup<any>
public step = 1;

[...]

ngOnInit(): void {
  this.pageForm = this.formBuilder.group({
    type: ['', Validators.required],
    content: [''],
    newForm: [false],
  });

[...]

  const type = this.pageForm.get(''type")?.value;
  //     ^   any
```

We can do this:
```ts
// === REACTIVE FORM ===
public pageForm = this.formBuilder.group({
  type: ['', Validators.required],
  content: [''],
  newForm: [false],
});
public step = 1;

[...]

  const type = this.pageForm.get(''type")?.value;
  //     ^   string | null | undefined
```

So this actually improves DX quite a lot.

There are also some cases where we can't build the form right at the property declaration (for example we need input data in order to build it), in those cases one approach I used in the past was putting it in a function and using ReturnType, you can check `context-datasource.component.ts` for example

PR #1416, opened 2023-06-01
feat/AB#54824_grouping-layers

matheus-relief:
# Description

This PR makes it possible to group map layer on the map settings.

## Ticket
https://dev.azure.com/WHOHQ/EMSSAFE/_boards/board/t/App%20Builder%20-%20WHE%20Dashboard/Stories?System.AssignedTo=%40me

## Type of change

# How Has This Been Tested?
See screenshots

## Sreenshots
Peek 2023-05-29 13-51

# Checklist:

( * == Mandatory ) 

### DIFF
--- a/libs/safe/src/i18n/en.json
+++ b/libs/safe/src/i18n/en.json
@@ -1180,6 +1180,7 @@
 					},
 					"position": "Position",
 					"series": {
+						"fill": "Fill",
 						"interpolation": {
 							"step": "Step interpolation",
 							"title": "Interpolation"
--- a/libs/safe/src/i18n/en.json
+++ b/libs/safe/src/i18n/en.json
@@ -409,11 +409,6 @@
 			"dashboard": {
 				"editFilter": "Edit filter",
 				"empty": "No content available. Wait for your administrator to build some.",
-				"filter": {
-					"empty": "No filters have been defined by the administrators",
-					"info": "Dashboard filters can be used between pages to filter their content.",
-					"title": "You are editing dashboard filters"
-				},
 				"filterSettings": "Filter settings",
 				"settings": {
 					"defaultPosition": "Default position for the filter"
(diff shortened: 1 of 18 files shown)


libs/safe/src/i18n/en.json
```diff
@@ -562,15 +557,6 @@
 			"subject": "Subject",
 			"withoutRecords": "Without records"
 		},
-		"emailPreview": {
```
AntoineRelief, line 565: @matheus-relief 
why did you remove that?

libs/safe/src/lib/components/widgets/map-settings/map-forms.ts
```diff
@@ -63,13 +63,17 @@ const DEFAULT_GRADIENT = [
  */
 export const createLayerForm = (value?: LayerModel) => {
   const type = get(value, 'type', 'FeatureLayer') as LayerType;
-  return fb.group({
+
+  return new FormGroup({
```
AntoineRelief, line 67: @matheus-relief 
please use form builder !

AntoineRelief: I'll merge it like that, but please explain me why you did this change

libs/safe/src/i18n/en.json
matheus-relief: Not on purpose, happened when I pulled from origin a while back and there were no translations for the other languages, so I just went with it, I can put it back and add the french keys if you want.

libs/safe/src/lib/components/widgets/map-settings/map-forms.ts
matheus-relief: Is there a reason why we should prioritize the form builder? 

But yeah, there's a difference in behavior (or maybe I just don't know how to use `fb` properly). I was having an issue with the sublayers. For some reason, when I used the fb, and there wasn't any sublayers when creating the form, the value for the sublayers was initialized as `null` instead of an empty array. I tried a bunch of things, including `fb.array`, but nothing really worked. So I just rewrote it using this syntax (that I'm more familiar with) and it solved the issue

matheus-relief: btw, I was using the same `get` function, with the `[]` as fallback, and still got `null`.

PR #1453, opened 2023-06-09
Feat/ab#66067 add avatar tinymce

AntoineRelief:
# Description

Added a way to add an avatar to the summary card.
The avatar is added to the summary card by adding a new field to the summary card settings. The field is a tagbox field that accepts a image urls.
The URL is then used to display the avatars in the summary card.
The user can set a limit to the number of avatars that can be displayed in the summary card.

Since Angular doesn't allow us to insert HTML directly, we have to use the DomSanitizer to sanitize the HTML before inserting it into the DOM,
and make the avatar group from scratch. There is probably a better way to do this, but this is the best I could come up with 😔.

The tinyMCE editor was updated with a new button to add the avatar group to the summary card.
The user can also add the avatar group by adding the following element to the summary card settings:
{{avatar.[name_of_tagbox_field] [width] [height] [limit]}}

The tinyMCE config was updated to created all the callback functions needed to add the avatar group to the summary card.
The config file is getting quite large, so I think we should refactor it a bit to make it more readable.

## Ticket

[Ticket [AB#66067](https://dev.azure.com/WHOHQ/2888b816-2d79-49ef-a86e-561e6cd456a0/_workitems/edit/66067)]

## Type of change

# How Has This Been Tested?

- [ ] Created a form with a tagbox with links to images. Created a summary card component in a dashboard, edited the cards using the new avatar button

## Sreenshots

image

image

image

# Checklist:

( * == Mandatory ) 


libs/safe/src/lib/components/widgets/summary-card/parser/utils.ts
```diff
@@ -239,6 +258,11 @@ const replaceRecordFields = (
         'gi'
       );
       formattedHtml = formattedHtml.replace(regex, convertedValue);
+      const regex2 = new RegExp(
```
AntoineRelief: @JulienKerlero 
is it needed?

PR #1561, opened 2023-06-27
Fix/ab#67729 filter selected item showing

ghost:
# Description

Fixed an issue where the previously selected item was showing in the option list when the list was filtered by the search input, even though it should've been filtered out.

To make this work, I had to change the way the <ui-select-menu> component was handling the optionList.onChanges event. Previously, it would set the label (displayTriggerText) to the corresponding item in the option list.
However, when the option list is **filtered**, the selected item **may not be in**, thus making the label empty.
If the selected value doesn't exist in the option list, the label should stay the same as the previous one, hopefully being the label of the selected item.
When the control is emptied, the selectedValues array is emptied as well, thus the label should be empty.

Known bug : 
When the selected value isn't in the first option list the <ui-select-component> tries to load, the label cannot be found (since it doesn't exist in the option list, and no previous label was set), therefore making the label empty.
Fixing this would require changing the way <ui-select-component> finds its label, or making an [input], which defies the purpose of it being able to find its label by itself.
Another solution would be, in the <safe-webmap-select>, to include the selected value as an additional option, if it's not in the default search result.

## Ticket

[Ticket 67729](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/67729)

## Type of change

# How Has This Been Tested?

- [ ] Created a map widget, opened the settings, selected a webmap, filtered the list, selected another webmap.

## Sreenshots

Selected option no longer show when filtered out
image

Video showing the tests and the known bug
https://github.com/ReliefApplications/oort-frontend/assets/60609610/b873cc58-733d-4b71-ab90-6a2b317db51c

# Checklist:

( * == Mandatory ) 


NathanHGit (commented):
It seems to work fine but it also creates a new bug.

peek

When you select a new option, search for text and delete it, the initialized option and the newly selected one appear.

Maximo-reliefapps (approved):
Seems good to me.

Fixed the bug where the searchbar became inaccessible, this was an issue with the select-menu component, where we had an `ng-if` hiding the `ng-content` when no element was in the list.
The issue with that is that we're passing the searchbar as part of the `ng-content`.
I removed that `ng-if`, checked other places using the same component and it doesn't seem to cause any issue.

Also changed the web-map component so it continues to show prev choices until new ones are loaded.

PR #1648, opened 2023-07-18
fix/survey-questions-readonly

estelafs:
# Description
Question isReadyOnly propriety never working, even if on a page without any logic configurations. Read only for questions only working when page also was in a read-only state
The issue didn't exist in the test survey but in the add a new record, update record, and update record in the grid modal

## Type of change

# How Has This Been Tested?
Creating questions that assume the readOnly state at different times and seeing them behave as they should.

## Screenshots
ro-add

ro-grid

# Checklist:

( * == Mandatory ) 

### DIFF
--- a/libs/safe/src/lib/components/form-modal/form-modal.component.ts
+++ b/libs/safe/src/lib/components/form-modal/form-modal.component.ts
@@ -232,6 +232,16 @@ export class SafeFormModalComponent
       this.temporaryFilesStorage
     );
 
+    // Set questions readOnly propriety
+    const structure = JSON.parse(this.form?.structure || '');
+    const pages = structure.pages;
+    pages.forEach((page: any) => {
+      page.elements?.forEach((question: any) => {
+        this.survey.getQuestionByName(question.name).readOnly =
+          question.readOnly ?? false;
+      });
+    });
+
     if (this.data.recordId && this.record) {
       addCustomFunctions(Survey, this.authService, this.record);
       this.survey.data = this.isMultiEdition ? null : this.record.data;
@@ -246,6 +256,16 @@ export class SafeFormModalComponent
     }
     this.survey.render(this.formContainer.nativeElement);
     // this.survey.render(this.containerId);
+    // this.pages.getValue().forEach((page: any) => {
+    //   console.log('page', page)
+    //   page.questions?.forEach((question: any) => {
+    //     const q = this.survey.getQuestionByName(question.name)
+    //     console.log('question',  q.name, question.readOnly, question.isReadOnly)
+    //     this.survey.getQuestionByName(question.name).readOnly = question.isReadyOnly ?? false;
+    //     console.log('question',  q.name, q.readOnly, q.isReadOnly)
+
+    //   });
+    // });
     this.loading = false;
   }
 
--- a/libs/safe/src/lib/components/form-modal/form-modal.component.ts
+++ b/libs/safe/src/lib/components/form-modal/form-modal.component.ts
@@ -236,7 +236,7 @@ export class SafeFormModalComponent
     const structure = JSON.parse(this.form?.structure || '');
     const pages = structure.pages;
     const initReadOnly = (elements: any): void => {
-      elements?.forEach((question: any) => {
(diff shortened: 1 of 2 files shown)


libs/safe/src/lib/components/form/form.component.ts
```diff
@@ -134,14 +134,12 @@ export class SafeFormComponent
...
-    if (!this.record) {
-      this.form.fields?.forEach((field) => {
-        if (field.readOnly) {
-          this.survey.getQuestionByName(field.name).readOnly = false;
-        }
-      });
-    }
+    // Set questions readOnly propriety
+    this.form.fields?.forEach((field) => {
+      if (field.readOnly) {
```
unai-reliefapp: Hi @estelafs , could happen that the question has readOnly state on and field not? In that case the false value wont be assigned, and also, readOnly property is always in the field, or only for the cases where is true? in order to take care of both values of readOnly i'll recommend to replace forEach content with this

```
      this.survey.getQuestionByName(field.name).readOnly = field.readOnly ?? false;
```
To cover all edge cases and dont leaving any loose case( as you did above).

KR, Unai

libs/safe/src/lib/components/form-modal/form-modal.component.ts
```diff
@@ -232,6 +232,27 @@ export class SafeFormModalComponent
...
+      elements?.forEach((question: any) => {
+        if (question.elements) {
+          // If question is a panel type that has sub-questions
+          initReadOnly(question.elements);
+        } else if (question.templateElements) {
+          // If question is a paneldynamic type that has sub-questions
+          initReadOnly(question.templateElements);
+        } else if (this.survey.getQuestionByName(question.name)) {
+          this.survey.getQuestionByName(question.name).readOnly =
+            question?.readOnly ?? false;
```
unai-reliefapp:
```suggestion
            question.readOnly ?? false;
```

libs/safe/src/lib/components/form-modal/form-modal.component.ts
```diff
@@ -232,6 +232,27 @@ export class SafeFormModalComponent
       this.temporaryFilesStorage
     );
 
+    // Set questions readOnly propriety
+    const structure = JSON.parse(this.form?.structure || '');
+    const pages = structure.pages;
+    const initReadOnly = (elements: any): void => {
+      elements?.forEach((question: any) => {
```
unai-reliefapp:
```suggestion
      elements.forEach((question: any) => {
```
You are already checking if elements exists before calling initReadOnly question, no need for duplicated checks 😜

libs/safe/src/lib/components/form-modal/form-modal.component.ts
```diff
@@ -232,6 +232,27 @@ export class SafeFormModalComponent
...
+          // If question is a paneldynamic type that has sub-questions
+          initReadOnly(question.templateElements);
+        } else if (this.survey.getQuestionByName(question.name)) {
+          this.survey.getQuestionByName(question.name).readOnly =
+            question?.readOnly ?? false;
+        }
+      });
+    };
+    pages.forEach((page: any) => {
+      initReadOnly(page.elements);
```
unai-reliefapp:
```suggestion
    if (page.elements) {
      initReadOnly(page.elements);
    }
```

unai-reliefapp (changes requested):
Just the comments I left, not critical, but this little PRs are perfect to train our attention to little details where the devil lives

unai-reliefapp (approved):

AntoineRelief:
@estelafs 

cc @unai-reliefapp 

I need to close this one, as it would not work:
- during addition, all fields should be editable ( if visible to the user ). It's a requirement from WHO. If we want to be more specific, and prevent that, we would need to add more options to surveyjs
- the logic would override some other logic we do in one of the methods of form-builder service. Indeed, depending on the permissions you have, some of the fields would be hidden / not editable. The code there would set as editable all fields that are still visible to the user, and could have been set as non editable by the createSurvey method of the service
should be fixed in this commit:
21b90101d354785afdf92ec64482e2c501487eb3

PR #1653, opened 2023-07-21
Ab#39346 infinite aggregations

MwanPygmay:
# Description

Front-end support for infinite aggregations. If a resource A references a resource B that references a resource C etc etc.., fields from resource C, D, E... should be available and selectable in aggregations. Circular dependency has been prevented, so if A references B that references A, fields from A and B should be available but fields from A should not be seen twice.
Names of the resource questions are displayed in blue with a little dropdown sign and can be expanded/collapsed upon clicking.
Nesting works only for group stages as it is the only stage that supports resource questions, even before this ticket => if needed, other stages are not too complicated to add.
Nested resources are shown with one level of indentation for each level of nesting.

## Ticket
[
Please insert link to ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/39346/)
[Link to back-end](https://github.com/ReliefApplications/oort-backend/pull/698)

## Type of change

# How Has This Been Tested?

Created a Parent form with ref to a resource A with ref to a resource B with ref to a resource C. Tried it on an aggregations in summary cards with a group field. Fields should be selectable. Resources questions can be expanded/collapsed (maybe this would also need to be rethought, especially for the mapping dropdown where expanding each of the fields one by one is just annoying). This should produce expected graph.

## Screenshots

infinite_aggregations

# Checklist:

( * == Mandatory ) 


libs/safe/src/lib/components/ui/aggregation-builder/pipeline/field-dropdown/field-dropdown.component.ts
```diff
@@ -1,6 +1,21 @@
...
+  AfterViewInit,
+  Component,
+  Input,
+  OnChanges,
+  QueryList,
+  SimpleChanges,
+  ViewChild,
+  ViewChildren,
+} from '@angular/core';
 import { UntypedFormControl } from '@angular/forms';
```
unai-reliefapp, line 11: @MwanPygmay Just a suggestion for the .html and .ts(not a requirement), as your a using a dot(`.`) as a separator for the nested properties name, i'll suggest to create a public property within the SafeFieldDropdownComponent class e.g.

`NESTED_FIELD_NAME_SEPARATOR = '.' as const`

So we have that value set in one place and if in the future we have to change that because the dot is not valid anymore(devs broke html language, we have to change how fields name behave and that includes the dot now, etc.) we just go there, change to the value we need a keep going 👍

libs/safe/src/lib/components/widgets/chart-settings/tab-main/tab-main.component.ts
```diff
@@ -127,6 +127,35 @@ export class TabMainComponent
       });
   }
 
+  /**
```
unai-reliefapp: If this function and the one in the aggregation-builder are the same let's try to move that to an auxiliary file and import this in both classes

Let's try to avoid code duplication at the bare minimum 👍 , app is big enough without duplications 😄 

(Maybe there is a refactor planned for this part of the aggregations already, but just in case)

unai-reliefapp (commented):
Check comments

libs/safe/src/lib/components/ui/aggregation-builder/pipeline/field-dropdown/field-dropdown.component.html
```diff
@@ -1,22 +1,73 @@
...
+    [isGroup]="true"
+    *ngIf="data.type.kind === 'LIST' || data.type.kind === 'OBJECT'"
+  >
+    <button
+      (click)="
+        toggleNodeExpansion(
+          parentName ? parentName + separator + data.name : data.name
+        )
+      "
+      class="min-h-[32px] flex items-center w-full text-left overflow-hidden cursor-pointer text-blue-400"
```
AntoineRelief, line 41: @MwanPygmay 
@matheus-relief 
no text blue !!
primary

AntoineRelief:
@MwanPygmay 
@matheus-relief 
closing this one, only because it's not mergeable in beta, but the code works 👍 
I duplicated in beta

PR #1665, opened 2023-07-26
Fix/ab#69615 last modified displaying wrong data

RenzoPrats:
# Description

In quick action buttons if the options "Pre-fill form with selected data" or "Attach to record" is selected and we also select "Auto-modify selected rows" when the button for the action is clicked even if the open modal is canceled the "Auto-modify selected rows" was happening, so now it only will happen when it's concluded the "Pre-fill form with selected data" or "Attach to record". Also modified the order of the options steps according to the comment below.

## Ticket

https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/69615

## Type of change

# How Has This Been Tested?

It has been tested verifying in the application if now the  "Auto-modify selected rows" only happen when it's concluded the "Pre-fill form with selected data" or "Attach to record".

## Screenshots

Peek 26-07-2023 17-03

# Checklist:

( * == Mandatory ) 

- [X] * The pull request is linked to an existing milestone


libs/safe/src/lib/components/widgets/grid/grid.component.ts
```diff
@@ -418,18 +418,44 @@ export class SafeGridWidgetComponent
...
       });
+
+      const savedSelectedRows = this.grid.selectedRows;
+
+      dialogRef.closed
+        .pipe(takeUntil(this.destroy$))
+        .subscribe((value: any) => {
+          if (value) {
+            if (this.blockModifySelectedRows && options.modifySelectedRows) {
+              this.promisedRowsModifications(
```
unai-reliefapp: Hi @RenzoPrats , why do we not wait to the row modification rows promise to end in order to trigger the reloadData method of the grid?

libs/safe/src/lib/components/widgets/grid/grid.component.ts
```diff
@@ -591,14 +619,34 @@ export class SafeGridWidgetComponent
...
+
+                          dialogRef.closed
+                            .pipe(takeUntil(this.destroy$))
+                            .subscribe((value: any) => {
+                              if (value) {
+                                if (
+                                  this.blockModifySelectedRows &&
+                                  options.modifySelectedRows
+                                ) {
+                                  this.promisedRowsModifications(
```
unai-reliefapp: Same as the previous comment

libs/safe/src/lib/components/widgets/grid/grid.component.ts
```diff
@@ -283,24 +286,21 @@ export class SafeGridWidgetComponent
...
-      await this.promisedRowsModifications(
-        options.modifications,
-        this.grid.selectedRows
-      );
-    }
 
     // Attaches the records to another one.
     if (options.attachToRecord && this.grid.selectedRows.length > 0) {
+      this.blockModifySelectedRows = true;
       await this.promisedAttachToRecord(
```
unai-reliefapp: Instead of so many params, maybe we can switch that to something like this

```
await this.promisedAttachToRecord(
this.grid.selectedRows,
options
);
```
As all data comes from `options` object, and we are sending the whole `options`  as other param in this change, it's redundant and less clear

And then extract those in the `promisedAttachToRecord` as this:
```
const {targetForm, targetFormField, targetFormQuery, modifySelectedRows, modifications } = options;
```

unai-reliefapp (changes requested):

AntoineRelief:
to be edited, so the steps are executed in this order:
- selectAll
- selectPage
- autoSave
- attachToRecord
- prefillForm
- modifySelectedRows
- notify
- publish
- sendMail
- goToNextStep | closeWorkflow

AntoineRelief:
@RenzoPrats 
in addition to what I put above:
attachToRecord -> should cancel the action if user clicks on cancel
prefillForm -> should cancel the action if user clicks on cancel

+ in the UI, when editing the action, please reorder the options, so they are ordered as described
+ in the UI, when editing the action, add a text above, telling that the steps will be executed in the displayed order

libs/safe/src/lib/components/widgets/grid/grid.component.ts
```diff
@@ -91,6 +92,8 @@ export class SafeGridWidgetComponent
     error: boolean;
   } = { error: false };
 
+  private blockModifySelectedRows = false;
```
AntoineRelief: @RenzoPrats 
it's not a correct approach
the code does not look well with this new parameter
I'll push new changes so you can compare

PR #1689, opened 2023-08-02
fix/AB#71534_remove-all-records-in-resources-question

matheus-relief:
# Description

The bug was caused because when getting the survey data, surveyJS considers empty array as undefined values, so it didn't send the data to the server in the first place. Thought there would be a configuration about that, but I don't think so, so I just go through all questions and if it's one with an array value and isn't present in the `survey.data`, add it as an empty array

## Useful links
* [SA-HQ- All signals - Resources data is not removing on update signal record](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/71534)
* [Forum post with an answer from SurveyJS support team](https://surveyjs.answerdesk.io/ticket/details/t10005/re-forcing-empty-value-to-be-included-in-survey-data)

## Type of change

# How Has This Been Tested?

By recreating the [steps to reproduce](https://dev.azure.com/WHOHQ/2888b816-2d79-49ef-a86e-561e6cd456a0/_apis/wit/attachments/3466dd91-def1-4b45-b8f7-33968885c352) on UAT

## Screenshots

Peek 2023-07-12 11-40

# Checklist:

( * == Mandatory ) 

### DIFF
--- a/libs/safe/src/lib/services/form-helper/form-helper.service.ts
+++ b/libs/safe/src/lib/services/form-helper/form-helper.service.ts
@@ -305,7 +305,6 @@ export class SafeFormHelpersService {
 
     // We want to include empty arrays multiple option questions
     survey.getAllQuestions().forEach((question) => {
-      console.log(question.getType());
       if (
         multipleOptionTypes.includes(question.getType()) &&
         question.visible
--- a/libs/safe/src/lib/services/form-helper/form-helper.service.ts
+++ b/libs/safe/src/lib/services/form-helper/form-helper.service.ts
@@ -291,4 +291,32 @@ export class SafeFormHelpersService {
 
     await Promise.all(promises);
   }
+
+  /**
+   * Gets survey data from the survey and adds empty arrays multiple option questions
+   *
+   * @param survey The survey to get the data from
+   * @returns The survey data
+   */
+  public getSurveyData(survey: Survey.SurveyModel): Record<string, any> {
+    const data = survey.data;
+
+    const multipleOptionTypes = ['checkbox', 'tagbox', 'resources'];
+
+    // We want to include empty arrays multiple option questions
+    survey.getAllQuestions().forEach((question) => {
+      console.log(question.getType());
+      if (
+        multipleOptionTypes.includes(question.getType()) &&
+        question.visible
+      ) {
+        const name = question.name;
+        if (!data[name]) {
+          data[name] = [];
+        }
+      }
+    });
+
+    return data;
+  }
 }
--- a/libs/safe/src/lib/components/form-modal/form-modal.component.ts
+++ b/libs/safe/src/lib/components/form-modal/form-modal.component.ts
@@ -383,7 +383,7 @@ export class SafeFormModalComponent
         mutation: EDIT_RECORD,
         variables: {
           id,
-          data: survey.data,
+          data: this.formHelpersService.getSurveyData(survey),
           template: this.data.template,
         },
       })
--- a/libs/safe/src/lib/components/form/form.component.ts
+++ b/libs/safe/src/lib/components/form/form.component.ts
@@ -271,7 +271,7 @@ export class SafeFormComponent
         mutation: EDIT_RECORD,
         variables: {
           id: recordId,
-          data: this.survey.data,
+          data: this.formHelpersService.getSurveyData(this.survey),
           template:
             this.form.id !== this.record?.form?.id ? this.form.id : null,
         },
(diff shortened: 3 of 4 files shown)


libs/safe/src/lib/services/form-helper/form-helper.service.ts
```diff
@@ -291,4 +291,32 @@ export class SafeFormHelpersService {
...
+   * @returns The survey data
+   */
+  public getSurveyData(survey: Survey.SurveyModel): Record<string, any> {
+    const data = survey.data;
+
+    const multipleOptionTypes = ['checkbox', 'tagbox', 'resources'];
+
+    // We want to include empty arrays multiple option questions
+    survey.getAllQuestions().forEach((question) => {
+      console.log(question.getType());
```
unai-reliefapp:
```suggestion
```
😄

unai-reliefapp: I was also wondering if instead of including this checks as a new method, include them in this already existing method, which is already correctly placed to set empty questions when needed
https://github.com/ReliefApplications/oort-frontend/blob/79c6f00a4ffd1254c3a1983bbae96609fbf59f9b/libs/safe/src/lib/services/form-helper/form-helper.service.ts#L68-L87

Not something to change, but as this method already exists, take advantage of it to set the empty questions value to all types

KR, Unai

unai-reliefapp (approved):

libs/safe/src/lib/services/form-helper/form-helper.service.ts
matheus-relief: Hey @unai-reliefapp, I actually tried that just now, but I couldn't get it to work. I changed the `setEmptyQuestions` code to be as follows

```ts
/**
 * Set data from survey empty questions
 *
 * @param survey Current survey to set up empty questions
 */
setEmptyQuestions(survey: Survey.SurveyModel): void {
  console.log(1);
  // We get all the questions from the survey and check which ones contains values
  const questions = survey.getAllQuestions();
  for (const field in questions) {
    const question = questions[field];
    if (question) {
      const key = question.getValueName();
      // If there is no value for this question
      if (!survey.data[key]) {
        // And is not boolean(false by default, we want to save that), we nullify it
        if (question.getType() !== 'boolean') {
          survey.data[key] = null;
        }
        // Or if is not visible or not actionable by the user, we don't want to save it, just delete the field from the data
        if (question.readOnly || !question.visible) {
          delete survey.data[key];
        }

        // Include empty arrays for multiple option questions if null
        const multipleOptionTypes = ['checkbox', 'tagbox', 'resources'];
        if (
          multipleOptionTypes.includes(question.getType()) &&
          question.visible
        ) {
          const name = question.name;
          if (!survey.data[name]) {
            survey.data[name] = [];
          }
        }
      }
    }
  }
}
```

I checked that it runs and that it goes into the if that does the `survey.data[name] = []` assignment, but then once we access `survey.data` again it's no longer there. I went to check and it makes sense as it's a setter/getter

```js
/**
 * Gets or sets an object with survey results. You can set this property with an object of the following structure:
 *
 * ```ts
 * {
 *   question1Name: question1Value,
 *   question2Name: question2Value,
 *   // ...
 * }
 * ```
 *
 * When you set this property in code, the new object overrides the old object that may contain default question values and entered data. If you want to *merge* the new and old objects, call the [`mergeData(newDataObj)`](https://surveyjs.io/form-library/documentation/surveymodel#mergeData) method.
 *
 * If you assign a new object while a respondent takes the survey, set the [`currentPageNo`](https://surveyjs.io/form-library/documentation/api-reference/survey-data-model#currentPageNo) property to 0 to start the survey from the beginning. This will also cause the survey to re-evaluate the [`visibleIf`](https://surveyjs.io/form-library/documentation/api-reference/question#visibleIf), [`enableIf`](https://surveyjs.io/form-library/documentation/api-reference/question#enableIf), and other [expressions](https://surveyjs.io/form-library/documentation/design-survey/conditional-logic#expressions).
 * @see setValue
 * @see getValue
 */
get data(): any;
set data(data: any);
```

unai-reliefapp: Ohhh, damn. Who knows what surveyjs does when retrieving data 😆 
Thanks for the very detail explanation @matheus-relief , then your solution is perfect 👍

AntoineRelief:
@matheus-relief 
closing this one
as we saw yesterday, the root cause of the issue affects more fields than just multi select
I deployed a new version on PROD with the issue fixed
you can see in the latest commits of next branch

PR #1701, opened 2023-08-07
fix/AB#71766-history-overflow-hidden

estelafs:
# Description
In the layout component, when we open the right sidenav and when the onResize event is detected, the new method setRightSidenavHeight() manually set the right sidenav max height and overflow proprieties to avoid hidden overflow issues, using the ui-sidenav-container height to calc the max-height available to the record History component

## Useful links

- Please insert link to ticket: [AB#71766 - ABC - History overflow hidden](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/71766)

## Type of change

# How Has This Been Tested?
Opening the history details of a record in the grid widget and using the scroll bar to see the complete history details without any overflow issue.

## Screenshots
historysidenav

# Checklist:

( * == Mandatory ) 

### DIFF
--- a/libs/ui/src/lib/sidenav/sidenav-container.component.html
+++ b/libs/ui/src/lib/sidenav/sidenav-container.component.html
@@ -4,9 +4,10 @@
   <ng-container *ngFor="let sidenav of uiSidenavDirective; let i = index">
     <div
       #sidenav
-      class="will-change-transform h-full overflow-y-auto bg-white translate-x-0"
+      class="will-change-transform overflow-y-auto bg-white translate-x-0"
       [attr.data-sidenav-show]="showSidenav[i]"
       [ngClass]="resolveSidenavClasses(i)"
+      [style.height]="height"
     >
       <div class="flex flex-col h-full"></div>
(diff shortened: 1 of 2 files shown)


matheus-relief (approved):

libs/safe/src/lib/components/layout/layout.component.ts
```diff
@@ -379,4 +388,23 @@ export class SafeLayoutComponent
     }
     return language;
   }
+
+  /**
+   * Manually set the right sidenav height to avoid hidden overflow issues.
+   */
+  setRightSidenavHeight(): void {
+    const recordHistory = this.document.getElementById('record-history');
```
AntoineRelief: @estelafs 
cc @matheus-relief 
it is too specific to the record history
we can indeed see the issue for the record-history, but the sidenavs are supposed to be generic

I think the easier would be to get all elements that use a specific id or a specific class, and to loop on them to set the height

AntoineRelief (changes requested):
too specific to history

PR #1704, opened 2023-08-08
Ability to filter users question by role

matheus-relief:
# Description

This PR adds the possibility to filter users question by role from the specified application

## Useful links
[Be able to filter user question by role](https://oortcloud.atlassian.net/jira/software/projects/OORT/boards/3?selectedIssue=OORT-658)

## Type of change

# How Has This Been Tested?

See bellow

## Screenshots
Peek 2023-07-12 11-40

# Checklist:

( * == Mandatory ) 

### DIFF
--- a/libs/safe/src/lib/survey/components/users.ts
+++ b/libs/safe/src/lib/survey/components/users.ts
@@ -72,7 +72,7 @@ export const init = (
                 value: r.id,
                 text: r.title?.split(' - ')[1],
               }));
-              choicesCallback(['', ...roles]);
+              choicesCallback([{value: null, text: ''}, ...roles]);
             });
           }
         },
--- a/libs/safe/src/lib/survey/components/users.ts
+++ b/libs/safe/src/lib/survey/components/users.ts
@@ -9,6 +9,7 @@ import {
   GetRolesFromApplicationsQueryResponse,
 } from '../graphql/queries';
 import { Question } from 'survey-angular';
+import { uniqWith } from 'lodash';
 
 /**
  * Inits the users component.
@@ -72,7 +73,7 @@ export const init = (
                 value: r.id,
                 text: r.title?.split(' - ')[1],
               }));
-              choicesCallback([{value: null, text: ''}, ...roles]);
+              choicesCallback([{ value: null, text: '' }, ...roles]);
             });
           }
         },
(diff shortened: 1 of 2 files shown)


libs/safe/src/lib/survey/components/users.ts
```diff
@@ -43,6 +57,27 @@ export const init = (
...
+        visibleIndex: 3,
+        choices: (obj: Question, choicesCallback: any) => {
+          if (obj.applications?.length) {
+            getApplicationRoles(obj.applications).subscribe(({ data }) => {
+              if (!data) choicesCallback([]);
+              const roles = data.rolesFromApplications.map((r) => ({
+                value: r.id,
+                text: r.title?.split(' - ')[1],
+              }));
+              choicesCallback(['', ...roles]);
```
unai-reliefapp:
```suggestion
              choicesCallback([{value: null, text: ''}, ...roles]);
```
We make sure all items follow same structure to avoid any unwanted behavior in all cases

libs/safe/src/lib/survey/components/users.ts
```diff
@@ -71,10 +106,18 @@ export const init = (
...
           if (data.users) {
-            const users: any = [];
+            const users: { value: string; text: string }[] = [];
             for (const user of data.users) {
-              if (!users.some((el: any) => el.value === user.id)) {
-                users.push({ value: user.id, text: user.username });
+              if (
+                user.id &&
+                user.username &&
+                !users.find((el) => el.value === user.id)
```
unai-reliefapp: In order to make this checks more performant, I'll avoid to use the `.find` check as it's going to be heavier as users arrays goes up in length. Instead of checking each time if the user is already added, maybe would be better to once the users array is finished, get all unique values at once using lodash for example:

`users = uniqWith(users, isEqual)` https://lodash.com/docs/4.17.15#uniqWith

Most of the time the duplicate cases would be minimal and we avoid a whole full array check logic and additional check for each user iteration from the data, what do you think?

unai-reliefapp (changes requested):
Changes to be checked

libs/safe/src/lib/survey/components/users.ts
matheus-relief: Makes a lot of sense, thanks for the tip!

unai-reliefapp (approved):

PR #1706, opened 2023-08-08
Bugfix/ab#69091 db layers edition still broken

unai-reliefapp:
# Description

refactor: change conditions style in order to be more readable and easy to debug for developers
fix: delete addLayer method on onAddLayer event in Layer class as that method only triggers when the layer is already added
fix: remaining listeners for each Layer instance trigger on switching between layers instance. Added method removeAllListeners to Layer class in order to clean up all listeners once we remove the related Layer class instance
fix: main duplication layer issue by always triggering layer editor form change updates once all related layers changes of the map are finished

## Useful links

- [Link to ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/69091)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Screenshots

# Checklist:

( * == Mandatory ) 


libs/safe/src/lib/components/ui/map/layer.ts
```diff
@@ -1023,4 +1042,20 @@ export class Layer implements LayerModel {
...
+   */
+  public removeAllListeners(map: L.Map) {
+    if (this.zoomListener) {
+      map.off('zoomend', this.zoomListener);
+    }
+    this.zoomListener = null as unknown as L.LeafletEventHandlerFn;
+    this.listeners.forEach((listener) => {
+      listener();
+    });
+    this.listeners = [];
```
matheus-relief, line 1077: I don't get what this does.

From what I understand, `this.listeners` is an array that will only ever contain `popupListeners`, why are we calling those functions (that pretty much just calls `this.popupService.setPopUp`) when removing all listeners?

libs/safe/src/lib/components/widgets/map-settings/edit-layer-modal/edit-layer-modal.component.ts
```diff
@@ -282,29 +300,40 @@ export class EditLayerModalComponent
...
+      this.form.valueChanges.pipe(startWith(this.form.value), pairwise()),
+      this.triggerFormChange.asObservable(),
+    ])
       .pipe(
-        startWith(this.form.value),
-        distinctUntilChanged((prev, next) => isEqual(prev, next)),
-        pairwise(),
+        filter(
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
+          ([[prev, next], trigger]) =>
```
matheus-relief, line 333:
```suggestion
          ([[, next]]) =>
```

libs/safe/src/lib/components/widgets/map-settings/edit-layer-modal/edit-layer-modal.component.ts
```diff
@@ -282,29 +300,40 @@ export class EditLayerModalComponent
...
+            this.triggerFormChange.getValue() &&
+            !isEqual(this.triggerFormChange.getValue(), next)
+        ),
         takeUntil(this.destroy$)
       )
       .subscribe({
-        next: ([prev, next]) => {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        next: ([[prev, next], trigger]) => {
```
matheus-relief:
```suggestion
        next: ([[prev, next]]) => {
```

libs/safe/src/lib/components/widgets/map-settings/edit-layer-modal/edit-layer-modal.component.ts
```diff
@@ -239,9 +245,13 @@ export class EditLayerModalComponent
 
   /**
    * Set default layer for editor
+   *
+   * @param lastFormValue Last value triggered from the layer editor form
    */
-  private setUpLayer() {
-    if (!this.data.mapComponent) return;
+  private setUpLayer(lastFormValue?: any) {
+    if (!this.data.mapComponent) {
```
matheus-relief, line 272: The issue where the legend is duplicated when you remove and add the geofiled again is still present. It happens when you zoom while the map is being updated

Peek 2023-07-12 11-40

_ignore the highlighted code, I thought I had fixed it but turns out that I didn't after more testing_

matheus-relief (changes requested):

libs/safe/src/lib/components/widgets/map-settings/edit-layer-modal/edit-layer-modal.component.ts
matheus-relief: I did some more testing and that does not fix the issue :disappointed:

matheus-relief: Seems like adding the following line to the beginning of `onRemoveLayer` fixes this issue. However, I'm not entirely sure it won't break something else

```ts
this.visibility = false;
```

libs/safe/src/lib/components/ui/map/layer.ts
unai-reliefapp: Hi @matheus-relief , absolutely right, I thought I used renderer for creating the listener, that's why that syntax(e.g. https://gist.github.com/Shijir/5e6b291747c39c7243bf4a3430b393db#file-renderer2-mousemove-document-ts-L19)

I'll update it right now, thanks

libs/safe/src/lib/components/widgets/map-settings/edit-layer-modal/edit-layer-modal.component.ts
unai-reliefapp: I get  your point, but even if they are not used would not be better to keep those so other devs know what's the context for each position?(maybe we have to work with prev or trigger in the future and we will know what and where are they)

libs/safe/src/lib/components/widgets/map-settings/edit-layer-modal/edit-layer-modal.component.ts
unai-reliefapp: mmm, you are right, if we zoom while the map is fetching the data from the server the onZoom triggers and as at that moment the map does not contain the layer, it adds it again:
```
if ( this.visibility &&
     !(layer as any).deleted &&

     *!map.hasLayer(layer)* 
) {
          map.addLayer(layer);
        }
```

I think that best solution would be to avoid any events while layer redraw from request is done, i'll double check that, thanks for spotting it!

libs/safe/src/lib/components/widgets/map-settings/edit-layer-modal/edit-layer-modal.component.ts
matheus-relief: Makes sense!

matheus-relief (approved):
lgtm!

libs/safe/src/lib/components/widgets/map-settings/edit-layer-modal/edit-layer-modal.component.ts
AntoineRelief: @unai-reliefapp 
cc @matheus-relief 
I think I still got an issue when editing layers..
you can see it while running the code with distant server, on app : 69091
I created a map with a group layer
the layers still got duplicated when zooming in / out during edition

I think it's because when you enter the group layer, you don't have any layer registered in the layers property of the map, so when you enter edition of one layer of this group, you don't unsubscribe to the events that would be registered on the layers of the group.
But if you exit the edition of a sublayer, the layers of the group are correctly registered, so when you enter again edition of a sublayer, you correctly unregister to the events of the group layers

taylordardono: @AntoineRelief Fix pushed in the last commit. 
Just one little thing, on updating the layer the backed returns the previous layer properties as seen in this image.
Data sent:
image
Data returned:
image

For now, on successful request Im just adding the properties sent to the backend in the map layers list item to reflect the change in the layer list, but I guess that the edit layer endpoint should return the layer with the latest properties. I can take a look at that in the backend and open the PR. In the meantime the fix added in this PR would handle any name update for the layers in the map layers list.

KR, Unai

unai-reliefapp: @AntoineRelief Fix pushed in the last commit. 
Just one little thing, on updating the layer the backed returns the previous layer properties as seen in this image.
Data sent:
image
Data returned:
image

For now, on successful request Im just adding the properties sent to the backend in the map layers list item to reflect the change in the layer list, but I guess that the edit layer endpoint should return the layer with the latest properties. I can take a look at that in the backend and open the PR. In the meantime the fix added in this PR would handle any name update for the layers in the map layers list.

KR, Unai

AntoineRelief (approved):

PR #1728, opened 2023-08-14
fix/AB#72417_right_sidenav_wrong_height

ghost:
# Description

Removed the height property on sidenav-container.component.ts (and its link to the html file) to fix the issue of the left sidenav being too short when opened.

Since the height property was computed from the parent's element height, it was omitting the navbar height, and the tailwind css 'h-full' was overwritten by the inline height property.

This issue is fixed in the 2.0.x by using the tailwind.config.js file 'important' property, but not in the 2.1.x.

## Useful links

- [Ticket 72417](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/72417)
- ~~Please insert link to back-end branch if any~~
- [Tailwind stance on the important rule](https://github.com/tailwindlabs/tailwindcss-typography/issues/26#issuecomment-659362241)

## Type of change

# How Has This Been Tested?

- [ ] Open a tab where the left sidenav shows. And size the windows down until it collapses. Open the sidenav.

## Screenshots

Works both in fullscreen (F11) mode and at any size when reshaping the window

https://github.com/ReliefApplications/oort-frontend/assets/60609610/ba0a2484-8842-43e2-b1ff-5d115f54ded3

# Checklist:

( * == Mandatory ) 


estelafs (approved):
Looks good to me (i checked that it didn't interfere with problematic things like the right sidenav or record history)!

unai-reliefapp:
> Looks good to me (i checked that it didn't interfere with problematic things like the right sidenav or record history)!

Hi, I double checked the styling sidenav on fullscreen and the height as full(100%) does not work because the parent component is absolute positioned, and if no inner content is set, the height is 0:
image
@JulienKerlero  Maybe instead of fully deleting the height you can use that attribute on fullscreen cases only.

You can subscribe to the browser's fullscreen property and if so, using renderer set the height attribute directly as it was done, and if there is no fullscreen, just the remove the height attribute directly.

You can place that logic here i guess: https://github.com/ReliefApplications/oort-frontend/blob/2.1.x/libs/ui/src/lib/sidenav/sidenav.directive.ts#L80-L83

Please double check that, any doubt feel free to ask
KR, Unai

unai-reliefapp (changes requested):
Please double check my comment

ghost:
@unai-reliefapp Do you mean this fullscreen button ? Because if so, even on the main 2.1.x branch, when in fullscreen, the sidenav doesn't show. Am I missing something ?
image

unai-reliefapp:
> @unai-reliefapp Do you mean this fullscreen button ? Because if so, even on the main 2.1.x branch, when in fullscreen, the sidenav doesn't show. Am I missing something ? image

Solved doubt with Julien

AntoineRelief:
@JulienKerlero 
I have to close this one
unfortunately, that would create another issue, with history
Indeed, the height of the right sidenav needs to be fixed, so the history can entirely be displayed
otherwise, you can get a difference of about 100px that would hide the last element of the list ( or more in some cases )

please see fix in these commits:
https://github.com/ReliefApplications/oort-frontend/commit/3192e074a8a1509d2695ad75e7ec4518dc069524
https://github.com/ReliefApplications/oort-frontend/commit/6e19aec629acb0eaaef617ba817e0b495d932c88

cc @estelafs

PR #1749, opened 2023-08-22
AB#73095  - Custom-styling-of-widgets-&-application-does-not-apply-to-text-widget-/-summary-card

AntoineRelief:
# Description

Now get the style as a string from the application Service. If the widget has a custom style, show the custom style, if it doesn't, show the global application style.

## Useful links

- [Ticket 73095](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/73095)
- ~~Please insert link to back-end branch if any~~
- ~~Please insert any useful link ( documentation you used for example )~~

## Type of change

# How Has This Been Tested?

- [ ] Created Text widget and Summary Cards with custom styles or global application style

## Screenshots

image

# Checklist:

( * == Mandatory ) 


libs/safe/src/lib/components/widgets/html-widget-content/html-widget-content.component.ts
```diff
@@ -12,7 +18,17 @@ import { SafeHtml } from '@angular/platform-browser';
...
 })
-export class HtmlWidgetContentComponent {
+export class HtmlWidgetContentComponent implements OnInit {
   @Input() html: SafeHtml = '';
   @Input() style?: string;
+
+  constructor(private elementRef: ElementRef) {}
+
+  ngOnInit(): void {
+    if (this.style) {
```
unai-reliefapp, line 33: With this change this style attribute won't be necessary right?
https://github.com/ReliefApplications/oort-frontend/blob/b5ed14cf03c7e6438be48eb25d39937aec75ca44/libs/safe/src/lib/components/widgets/html-widget-content/html-widget-content.component.html#L4

libs/safe/src/lib/components/widgets/editor/editor.component.ts
```diff
@@ -59,6 +59,7 @@ export class SafeEditorComponent implements OnInit {
   /** Sanitizes the text. */
   ngOnInit(): void {
     this.setContentFromLayout();
+    this.formattedStyle = this.settings.widgetDisplay.style;
```
unai-reliefapp:
```suggestion
    this.formattedStyle = this.settings.widgetDisplay?.style;
```
For empty text widgets:
image

unai-reliefapp:
@JulienKerlero @AntoineRelief  The settings shown below are not set in the text widget until the user opens the settings in the menu dropdown and saves anything.

If you try to style the widget or do any kind of action without saving first a setting config in the text widget it would break and never be applied because those settings (widgetDisplay among others) are missing and only set on creating the tileForm

image

I'll recommend to add those settings in the dashboard model for the text widget by default if we can use them from the begining

unai-reliefapp:
@JulienKerlero Check again because the last commit does not solve the issue, as I said in the previous comment the best way would be to update this line in here https://github.com/ReliefApplications/oort-frontend/blob/2.1.x/libs/safe/src/lib/models/dashboard.model.ts#L180
With a default value of widgetDisplay:

```
  widgetDisplay: {
        showBorder: true,
        style: '',
      },
```

Also take in account that all text widgets saved previous to this change if they dont have any settings saved they would fail because application uses the config saved in the data base, therefor you'll also have to add a condition to check if the loaded text widget contains a `widgetDisplay` property within, and if not, set it with the default value

Let me know if there is any doubt
KR, Unai

AntoineRelief:
@TaiKamilla 
you can continue the work on this one
I see two issues:
- widget style not applied for summary card ( at least from what I tested )
- when you edit the style, you don't dynamically update the style

TaiKamilla:
Update the live preview
Here is a video

https://github.com/ReliefApplications/oort-frontend/assets/15035750/39244ac5-8a1b-48d4-bcb6-f5a23dc52987

AntoineRelief:
@TaiKamilla 
ℹ️  don't start working on that, I'm looking at it
issues I detected so far:
- the page sometimes refreshes for now obvious reason, hiding all the content except a notification to indicate that the dashboard or application has been updated

apps/back-office/src/app/components/custom-style/custom-style.component.ts
```diff
@@ -100,19 +100,7 @@ export class CustomStyleComponent
   }
 
   ngOnInit(): void {
-    if (this.applicationService.rawCustomStyle) {
```
AntoineRelief, line 103: @TaiKamilla 
why did you remove that?

libs/safe/src/lib/components/widgets/html-widget-content/html-widget-content.component.ts
```diff
@@ -12,7 +18,22 @@ import { SafeHtml } from '@angular/platform-browser';
   styleUrls: ['./html-widget-content.component.scss'],
   encapsulation: ViewEncapsulation.ShadowDom,
 })
-export class HtmlWidgetContentComponent {
+export class HtmlWidgetContentComponent implements OnInit {
   @Input() html: SafeHtml = '';
   @Input() style?: string;
+
+  /**
+   * Constructor for safe-html-widget-content component
```
AntoineRelief: @TaiKamilla @JulienKerlero 
I didn't check who wrote that, but you don't want to indicate in a constructor jsdoc that it is a constructor
the comment there is what you would display when you hover the component in another file for example, so seeing "constructor" while you would need to understand what the file is about does not help
it should be the same text than the class jsdoc

apps/back-office/src/app/components/custom-style/custom-style.component.ts
TaiKamilla: I think this was cos if you leave this, it will write the style in the header, and is not necesary
but maybe I deleted to much?

libs/safe/src/lib/components/widget/widget.component.ts
```diff
@@ -55,38 +44,4 @@ export class SafeWidgetComponent implements OnInit, OnDestroy {
...
-  /**
-   * Widget component
-   *
-   * @param restService Shared rest service
-   */
-  constructor(private restService: SafeRestService) {}
-
-  ngOnInit(): void {
-    // Get style from widget definition
-    const style = get(this.widget, 'settings.widgetDisplay.style') || '';
```
AntoineRelief, line 68: @TaiKamilla 
I think you also removed there the possibility for other widgets than text & summary cards to be styled

AntoineRelief:
@TaiKamilla 
cc @JulienKerlero 
unfortunately this ticket is removing too many features:
- we cannot save anymore scss in widgets
- application style is not editable anymore
- widget style only applies to text & editor widgets..

PR #1754, opened 2023-08-23
OORT-fix/IM-8-date-format-change-by-preference

estelafs:
# Description
The date and time format chosen in the preferences updates the date format in the platform, but not in the date picker in the surveyjs for single input questions with date, datetime-local or time types.

## Useful links

- Please insert link to ticket: [Date format doesnt change by preference](https://oortcloud.atlassian.net/browse/IM-8)

## Type of change

# How Has This Been Tested?
Changing the date and time format in the preferences before add/update a record with date or time questions.

## Screenshots
fr format:
fr

en format:
en

# Checklist:

( * == Mandatory ) 

### DIFF
--- a/libs/safe/src/lib/survey/widgets/text-widget.ts
+++ b/libs/safe/src/lib/survey/widgets/text-widget.ts
@@ -10,6 +10,7 @@ import { JsonMetadata, SurveyModel } from 'survey-angular';
 import { Question, QuestionText } from '../types';
 import { ButtonComponent } from '@oort-front/ui';
 import { IconComponent } from '@oort-front/ui';
+import { SafeDateTranslateService } from '../../services/date-translate/date-translate.service';
 
 type DateInputFormat = 'date' | 'datetime' | 'datetime-local' | 'time';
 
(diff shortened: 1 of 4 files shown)


libs/safe/src/lib/survey/widgets/text-widget.ts
```diff
@@ -85,7 +91,8 @@ export const init = (Survey: any, domService: DomService): void => {
             pickerDiv = document.createElement('div');
             const pickerInstance = createPickerInstance(
               question.inputType as DateInputFormat,
-              pickerDiv
+              pickerDiv,
+              dateTranslate.currentLang
```
unai-reliefapp: Hi @estelafs , watch out for this use of the service and property, because preferences gives you the option to specify language and format independantly, which means that currentLang is not always a valid value as it only refers to the selected language, not format.

Here I have `fr` language and `en` format
image

1.  If is not the expected behavior to keep both properties loosely coupled from each other, first we should fix that.
2. If it's the expected behavior, then you'll have to use the localStorage property `date-lang` which contains the exact language for the format, as you can see in the service
https://github.com/ReliefApplications/oort-frontend/blob/2.1.x-oort/libs/safe/src/lib/services/date-translate/date-translate.service.ts#L37
3. Also the terniary added for each type of format actually is good but once we add a new language(which could happen) you'll have to rewrite those `currentLang === 'fr' ? 'HH:mm' : 'hh:mm a';`. I'll recommend to add a new file containing a key-value pair of formats for each language, and get the `date-lang` from the localStorage and always set a default base case and then easily get the format from that new file

```
let currentFormatLang = localStorage.get('date-lang') ?? 'en';
...
datePickerInstance.format = keyValueFormatObject[currentFormatLang];
```
instead of passing by the service across multiple files to keep these as loosely coupled as possible and to avoid huge property handle that would eventually lead to errors and or at least to a less clear code 😄 


Let me know what you think
KR, Unai

unai-reliefapp (changes requested):

unai-reliefapp (approved):
Great set!

PR #1760, opened 2023-08-25
Feat 2.0.x/ab#73217 UI table elements max width

NathanHGit:
# Description

Set a max-width on some ui table elements. When the text is too long, it will be cropped and "..." will be added at the end.

## Useful links

Ticket: [73217 - ABC - Improve ui-table elements so the text fields have a max-width](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/73217/)

## Type of change

# How Has This Been Tested?

I checked each changed table to see if the new max-width was correct.

## Screenshots

image

# Checklist:

( * == Mandatory ) 


Maximo-reliefapps (approved):
Working well, moved the css max-width property to tailwind.
I created a new class on tailwind.config to avoid having arbitrary value classes on so many elements ( max-w-[30vw] => max-w-30 )

unai-reliefapp:
Hi @NathanHGit not only title or name cells are the ones that could have long content
For example in the subscriptions table, converTo column also could have

image
_Capture taken from this branch_

My recommendation, as all cells are using our uiCell directive: https://github.com/ReliefApplications/oort-frontend/blob/2.0.x/libs/ui/src/lib/table/cell.directive.ts#L32
You can add Maxi's class there directly and would be automatically update to all columns of our tables. Test some cases and let us know if it's easier like this 👍 

KR, Unai

tailwind.config.js
```diff
@@ -46,6 +46,9 @@ module.exports = {
         fadeIn: 'fadeIn 0.5s forwards',
         fadeOut: 'fadeOut 0.1s forwards',
       },
+      maxWidth: {
```
AntoineRelief: @NathanHGit 
we don't need that in fact
tailwind allows to create classes on the go

libs/safe/src/lib/style/styles.scss
```diff
@@ -161,6 +161,11 @@ body {
   overflow: hidden;
 }
 
+td {
```
AntoineRelief: @NathanHGit 
we should not put that there, but at the component level

AntoineRelief:
@NathanHGit @Maximo-reliefapps 
ticket was resolved but the code was incorrectly written
I invite you both to have a look at the recent updates of the ticket:
- we don't need to update tailwind config in most cases
- we should have applied the overflow properties at the component level, not at the root level

unai-reliefapp:
> Hi @NathanHGit not only title or name cells are the ones that could have long content For example in the subscriptions table, converTo column also could have
> 
> image _Capture taken from this branch_
> 
> My recommendation, as all cells are using our uiCell directive: https://github.com/ReliefApplications/oort-frontend/blob/2.0.x/libs/ui/src/lib/table/cell.directive.ts#L32 You can add Maxi's class there directly and would be automatically update to all columns of our tables. Test some cases and let us know if it's easier like this 👍
> 
> KR, Unai

@AntoineRelief This was not addressed, just to keep in mind

PR #1822, opened 2023-09-11
feat/AB#74580_add_pagination_main_users_in_backoffice

RenzoPrats:
# Description

Added pagination to main users table in back-office.

## Useful links

[ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/74580)
[backend](https://github.com/ReliefApplications/ems-backend/pull/734)

## Type of change

# How Has This Been Tested?

It has been tested accessing the users table in the application and testing if the pagination is working correctly.

## Screenshots

Peek 12-09-2023 17-14

# Checklist:

( * == Mandatory ) 


apps/back-office/src/app/dashboard/pages/users/users.component.ts
```diff
@@ -24,31 +34,133 @@ export class UsersComponent implements OnInit {
...
-          .watchQuery<GetRolesQueryResponse>({
-            query: GET_ROLES,
+    this.usersQuery = this.apollo.watchQuery<GetUsersQueryResponse>({
+      query: GET_USERS,
+      variables: {
+        first: ITEMS_PER_PAGE,
+        afterCursor: null,
+      },
+    });
+    this.usersQuery.valueChanges.subscribe((resUsers) => {
```
unai-reliefapp: Hi @RenzoPrats , now that we are adding new query subscriptions, please add the unsubscribe teardown logic for those(SafeUnsubscribeComponent),

```suggestion
    this.usersQuery.valueChanges.pipe(takeUntil(this.destroy$).subscribe((resUsers) => {
```

Same for the query of roles.
Thanks

unai-reliefapp (changes requested):

apps/back-office/src/app/dashboard/pages/users/users.component.ts
RenzoPrats: Hi @unai-reliefapp, I fixed it!
Thanks

estelafs (changes requested):
Table pagination isn't updated when adding or deleting users and the filter only considers the visible contents of the table instead of refetching using the filters params. "On page changes" event don't update the pagination number after deleting, only after adding user. Requested changes in the backend as well.

RenzoPrats:
@estelafs I fixed these issues, if you can please review it again!

estelafs (approved):

unai-reliefapp:
@AntoineRelief Now that we add pagination feature to the user list, filter should be applied in the query now right? Not filtering current users in the front but an argument sent into the new paginated query

KR,  Unai

apps/back-office/src/app/dashboard/pages/forms/forms.component.ts
```diff
@@ -232,6 +232,11 @@ export class FormsComponent extends SafeUnsubscribeComponent implements OnInit {
                   (x) =>
                     x.id !== form.id && form.id !== x.resource?.coreForm?.id
                 );
+                // clear(evict) forms query in cache
+                this.apollo.client.cache.evict({
+                  id: 'ROOT_QUERY',
+                  fieldName: 'forms',
+                });
```
Matthis-M-ReliefApps, line 237: What's the purpose of that?

apps/back-office/src/app/dashboard/pages/pull-jobs/pull-jobs.component.ts
```diff
@@ -92,7 +92,7 @@ export class PullJobsComponent
       query: GET_PULL_JOBS,
```
Matthis-M-ReliefApps, line 90: Why the changes on pulljobs? Is that linked to another ticket?

libs/safe/src/lib/components/users/users.component.ts
```diff
@@ -308,12 +285,15 @@ export class SafeUsersComponent
   applyFilter(column: string, event: any): void {
     if (column === 'role') {
       this.roleFilter = event?.trim() ?? '';
-    } else {
+      this.changeFilter.emit({ column: column, event: this.roleFilter });
+    } else if (column === 'search') {
       this.searchText = event
         ? event.target.value.trim().toLowerCase()
```
Matthis-M-ReliefApps: Noticed that this is causing caps to go lowercase on the UI also, could you try to fix it also?
Peek 27-10-2023 11-45

apps/back-office/src/app/dashboard/pages/users/users.component.ts
```diff
@@ -18,37 +24,248 @@ import {
...
+   * Change page length on invite or delete users
+   *
+   * @param e lenght of users added.
+   */
+  public changePageLength(e: any) {
+    if (e.op === 'add') {
+      if (this.cachedUsers.length === this.pageInfo.length) {
+        e.data.forEach((usr: any) => {
+          this.cachedUsers = this.cachedUsers.concat([usr]);
+        });
```
Matthis-M-ReliefApps: Is there a reason to do this instead of `this.cachedUsers = this.cachedUsers.concat(e.data)` ?

apps/back-office/src/app/dashboard/pages/resources/resources.component.ts
```diff
@@ -224,6 +224,11 @@ export class ResourcesComponent
               this.resources = this.resources.filter(
```
Matthis-M-ReliefApps, line 222: Why those changes here also?

apps/back-office/src/app/dashboard/pages/users/users.component.html
```diff
@@ -1 +1,18 @@
-<safe-users [loading]="loading" [users]="users" [roles]="roles"></safe-users>
```
Matthis-M-ReliefApps, line 1: I think safe-users is used only on this page, is there a reason not to implement the paginator directly inside the <safe-users> component? I feel like it could've been easier no?

apps/back-office/src/app/dashboard/pages/users/users.component.ts
```diff
@@ -18,37 +24,248 @@ import {
...
+      this.users = this.cachedUsers.slice(
+        ITEMS_PER_PAGE * this.pageInfo.pageIndex,
+        ITEMS_PER_PAGE * (this.pageInfo.pageIndex + 1)
+      );
+    }
+    // clear(evict) users query in cache
+    this.apollo.client.cache.evict({
+      id: 'ROOT_QUERY',
+      fieldName: 'users',
+    });
```
Matthis-M-ReliefApps, line 206: It's the first time "this.apollo.client.cache.evict" is used in the code, can you explain what's its purpose?

apps/back-office/src/app/dashboard/pages/users/users.component.ts
```diff
@@ -18,37 +24,248 @@ import {
...
+            value: e.event,
+          });
+        }
+      }
+    } else {
+      // clear filters
+      this.filter.filters = [];
+    }
+    this.fetchUsers(true);
+  }
```
Matthis-M-ReliefApps: This seems to work correctly, but appears a bit too complicated to me. Would it be possible to restructure the code so the filters can be used as they are in _apps/back-office/src/app/dashboard/pages/applications/applications.component.html_
The filter is built and updated in app-filter:
image
Then emitted as an event:
image
And directly used by the applications component like this:
image
Which makes the whole flow way simpler. Could you check if that's possible in our case too? But if it's actually not possible or makes it more complicated, don't bother.

apps/back-office/src/app/dashboard/pages/users/users.component.ts
```diff
@@ -18,37 +24,248 @@ import {
...
           });
       });
   }
+
+  /**
+   * Handles page event.
+   *
+   * @param e page event.
+   */
+  onPage(e: UIPageChangeEvent): void {
```
Matthis-M-ReliefApps: The logic seems right! However, could you explain why implement all this this way instead of using the same logic as in other components? (libs/safe/src/lib/components/notifications/notifications.component.ts, apps/back-office/src/app/dashboard/pages/resources/resources.component.ts, libs/safe/src/lib/views/application-users/components/user-list/user-list.component.ts, ...)

libs/safe/src/lib/components/users/components/add-user/add-user.component.ts
```diff
@@ -1,20 +1,15 @@
-import { Component, Inject, OnInit } from '@angular/core';
```
Matthis-M-ReliefApps, line 1: To explain the change on this file:
This is used when adding new users to the platform or application
I think there shouldn't be autocompletion here for two reasons:
1. When you're inviting from inside an application, even if you have rights to manage users inside the application, you're not supposed to see users outside of it, so you shouldn't have access to their emails, even as autocomplete
2. When you're inviting from outside an application, the users you want to add are, by definition, not in the system yet, so an autocomplete will never be possible.
Hence, I don't see a use case where the autocomplete is useful here, but please correct me if I'm wrong!

Matthis-M-ReliefApps (changes requested):
Feels good to finally have pagination on this users list :+1: Especially with the new app that makes it grow to 5k users!
I took my time to review with a large scope so I have a bunch of questions

apps/back-office/src/app/dashboard/pages/forms/forms.component.ts
RenzoPrats: Hi @Matthis-M-ReliefApps the purpose of that is clear the cache, because when we add/remove some user and then we click in an item and then come back to the pagination we enter in something like this 
```
 if (cachedValues) {
      this.updateValues(cachedValues, false);
```
and we does not see the right pagination(not added the new user or not removed the user)

apps/back-office/src/app/dashboard/pages/pull-jobs/pull-jobs.component.ts
RenzoPrats: It's because the pagination of the users was did based on the pagination of the pulljobs and I noticed an issue there, so I fixed it in the users and then in pulljobs

libs/safe/src/lib/components/users/users.component.ts
RenzoPrats: Yes, sure!

apps/back-office/src/app/dashboard/pages/users/users.component.ts
RenzoPrats: No reason, we can change to this.cachedUsers = this.cachedUsers.concat(e.data)

apps/back-office/src/app/dashboard/pages/users/users.component.ts
RenzoPrats: Do you mean using handleTablePageEvent? I think we changed the logic since I did this PR.

apps/back-office/src/app/dashboard/pages/users/users.component.html
RenzoPrats: In fact, safe-users is used only on this page, so there's no specific reason for implement like this.

apps/back-office/src/app/dashboard/pages/resources/resources.component.ts
RenzoPrats: For delete the user when is clicked to delete an user.

apps/back-office/src/app/dashboard/pages/users/users.component.ts
RenzoPrats: Yes, it's possible but we need first to change the pagination to safe-users.

libs/safe/src/lib/components/users/users.component.ts
```diff
@@ -121,20 +279,46 @@ export class SafeUsersComponent
...
-                    this.translate.instant('components.users.onInvite.plural')
-                  );
+                let alreadyInvited = false;
+                // verify if users have already been invited
+                this.users.forEach((usr: any) => {
+                  data?.addUsers.forEach((usrData: any) => {
+                    if (usrData.username === usr.username) {
+                      alreadyInvited = true;
+                    }
+                  });
```
Matthis-M-ReliefApps: It works but I think you can try to improve it using lodash:
https://lodash.com/docs/4.17.15#intersection
Like get the intersection and check if there is at least one element in it, let me know if I'm not clear!

libs/safe/src/lib/components/users/users.component.ts
```diff
@@ -121,20 +279,46 @@ export class SafeUsersComponent
...
+                    this.snackBar.openSnackBar(
+                      this.translate.instant('components.users.onInvite.plural')
+                    );
+                  } else {
+                    this.snackBar.openSnackBar(
+                      this.translate.instant(
+                        'components.users.onInvite.singular'
+                      )
+                    );
+                  }
```
Matthis-M-ReliefApps: Since all the snackbars are the same, I think you can just change the message and error? Like:

```
let messagePath = '';
let isError = false;
if (cond1) {
messagePath = "mess1"
} else if (cond2 {
messagePath = "mess2";
isError = true;
} else ...
}
this.snackBar.openSnackBar(this.translate.instant(messagePath), error: error);
```

It should clear a few lines! (Sorry I'm being super picky :laughing:)

apps/back-office/src/app/dashboard/pages/forms/forms.component.ts
Matthis-M-ReliefApps: Thanks for the explanation Renzo! I don't have the expertise on Apollo so I will ask for another pair of eyes :laughing: 

Hi @unai-reliefapp ! Antoine advised me to check with you:
Since this is an Apollo functionality is present nowhere else in the code and we would add it, would you have a minute to dig on this and share your opinion? I'm a bit wary of any unexpected change in Apollo behaviour...
Thanks!!

apps/back-office/src/app/dashboard/pages/users/users.component.html
Matthis-M-ReliefApps: The new implementation looks good! Thanks!

Matthis-M-ReliefApps:
I noticed another issue while switching pages after changing the amount of users fetched, sometimes with 50 users set, the table only displays 10. Here's a GIF to demonstrate it:

Peek 31-10-2023 16-52

apps/back-office/src/app/dashboard/pages/users/users.component.ts
Matthis-M-ReliefApps: Would something like this work in our case also?
image

Seems quite simpler, but maybe I'm missing something?

Also, if you compare the length of cachedData and received data, what happens if we remove then add a user? Is the cache still updated or not?

libs/safe/src/lib/components/utils/update-queries.ts
Matthis-M-ReliefApps, line 1: I think it's a good idea to regroup this in "libs/safe/src/lib" :+1: 

There seems to be 4 mostly identical files in the end :laughing: 
image
Can you also remove the other duplicates in "back-office" so that's the only version? (to avoid duplication)

If that causes an import/build issue or anything else, please let me know.

Also, I recommend doing a merge from 2.x.x before this change to minimize the risks of conflicts.

Matthis-M-ReliefApps (changes requested):
Some questions/changes still but it's looking better already! :+1:

apps/back-office/src/app/dashboard/pages/users/users.component.ts
RenzoPrats: Done, indeed looks better!

libs/safe/src/lib/components/users/users.component.ts
RenzoPrats: Improved using lodash!

libs/safe/src/lib/components/users/users.component.ts
RenzoPrats: Done! Don't worry, I agree with these changes in fact looks better!

libs/safe/src/lib/components/utils/update-queries.ts
RenzoPrats: Removed duplicated update-queries, the only issue I had was with the update-queries from ui, I were not able to remove it otherwise I get circular dependency problem.

RenzoPrats:
> I noticed another issue while switching pages after changing the amount of users fetched, sometimes with 50 users set, the table only displays 10. Here's a GIF to demonstrate it:
> 
>  [ Peek 31-10-2023 16-52 ](https://user-images.githubusercontent.com/80319175/279437585-293b7379-7b1a-4a4f-a8af-bb475ed06f69.gif) [ ](https://user-images.githubusercontent.com/80319175/279437585-293b7379-7b1a-4a4f-a8af-bb475ed06f69.gif)

I was not able to reproduce this issue, I think it's working fine after I changed the onPage function as proposed in these comments.

RenzoPrats:
@Matthis-M-ReliefApps  I only found one more issue, when we add new users we should not add if it's already been added and display a message. The problem is that in the frontend we have two lists users and cachedUsers, the cachedUsers will store all the users that we already loaded but if we want to add an user that we already have but we didn't load yet we will be allowed, and in the backend we have the possibility to update the existing ones so we cannot return an error if we already have this user, so the only solution in this desing is frontend but as we didn't load everything from the beginning we cannot do it in the frontend too.

MwanPygmay (commented):
Tested, read code.
There was a problem: no sorting on data implied mongo natural order, which is not sorted and resulted in lacks of data (because ids in mongo can make [1,2,3,5,4,6], and thus we could store the 5 as a cursor, then missing the 4). Fixed it by adding a sorting on created at (maybe name would be more coherent?)
Otherwise, looks good to me

Matthis-M-ReliefApps (approved):
I have a feeling that there could be some improvements to make on the way we use apollo caching/pagination/filtering (specifically, make better use of apollo existing cache system instead of using our own cachedUsers on the users.component), but aside from this it seems to be working well!

libs/safe/src/index.ts
```diff
@@ -113,3 +113,4 @@ export * from './lib/pipes/cron-parser/public-api';
 
 // === UTILS ===
 export * from './lib/utils/public-api';
+export * from './lib/utils/update-queries';
```
AntoineRelief, line 116: check line above, there was already a file meant to export code from utils folder

AntoineRelief:
closing and putting more of changes + some additional ones in this pr:
https://github.com/ReliefApplications/ems-frontend/pull/2147

PR #1838, opened 2023-09-15
Fix/ab#75167 prefill user data after saved one

RenzoPrats:
# Description

Fixed prefill user data when using "new record" after having saved one.

## Useful links

[ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/75167)

## Type of change

# How Has This Been Tested?

It has been tested creating a form with some fields with default value expressions as described in the ticket and adding a record and then clicking in "new record" and verifying if the fields were prefilled.

## Screenshots

Peek 15-09-2023 17-02

# Checklist:

( * == Mandatory ) 

### DIFF
--- a/libs/safe/src/lib/components/form/form.component.ts
+++ b/libs/safe/src/lib/components/form/form.component.ts
@@ -164,11 +164,15 @@ export class SafeFormComponent
         //verify if have defaultValueExpression
         if (question.defaultValueExpression) {
           //get the defaultExpression in the right format
-          const questionDefaultExpression = question.defaultValueExpression.replace('{', '').replace('}', '');
+          const questionDefaultExpression = question.defaultValueExpression
+            .replace('{', '')
+            .replace('}', '');
           //verify if the defaultExpression is setted in variables
           if (this.survey.getVariable(questionDefaultExpression)) {
             //update the question value with the variable
-            question.updateValueFromSurvey(this.survey.getVariable(questionDefaultExpression));
+            question.updateValueFromSurvey(
+              this.survey.getVariable(questionDefaultExpression)
+            );
           }
         }
       });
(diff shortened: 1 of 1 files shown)


libs/safe/src/lib/components/form/form.component.ts
```diff
@@ -157,6 +157,25 @@ export class SafeFormComponent
 
     if (cachedData) {
       this.survey.data = cachedData;
+      this.formHelpersService.addUserVariables(this.survey);
```
unai-reliefapp: ey @RenzoPrats Could you please move this new feature into a single function inside the class to avoid code duplication?
Thanks

unai-reliefapp (changes requested):

libs/safe/src/lib/components/form/form.component.ts
RenzoPrats: @unai-reliefapp sure, it's done.
Thanks.

unai-reliefapp (approved):

libs/safe/src/lib/components/form/form.component.ts
```diff
@@ -157,6 +157,21 @@ export class SafeFormComponent
...
+      const surveyQuestion = this.survey.getAllQuestions();
+      //get all questions
+      surveyQuestion.forEach((question: any) => {
+        //verify if have defaultValueExpression
+        if (question.defaultValueExpression) {
+          //get the defaultExpression in the right format
+          const questionDefaultExpression = question.defaultValueExpression.replace('{', '').replace('}', '');
+          //verify if the defaultExpression is setted in variables
```
AntoineRelief: @RenzoPrats 
I'm seeing you're doing that whenever you comment
please leave a space between '//' and the text

libs/safe/src/lib/components/form/form.component.ts
```diff
@@ -205,11 +206,37 @@ export class SafeFormComponent
...
   }
 
+  /**
+   * Set user variables
+   */
+  private setUserVariables(): void {
+    this.formHelpersService.addUserVariables(this.survey);
+    const surveyQuestion = this.survey.getAllQuestions();
+    //get all questions
+    surveyQuestion.forEach((question: any) => {
```
AntoineRelief: @RenzoPrats 
aware of the naming
there, this.survey.getAllQuestions return an array, but you're using singular for the const
make sure to prefer plural in that case

libs/safe/src/lib/components/form/form.component.ts
```diff
@@ -205,11 +206,37 @@ export class SafeFormComponent
     // });
   }
 
+  /**
+   * Set user variables
```
AntoineRelief: @RenzoPrats 
the comment there is not enough
it's exactly the name of the method, so we already have this information from the method name
you should explain more the logic, tell what the method does

PR #1839, opened 2023-09-15
feat/AB#75138-reposition-save-and-cancel-buttons

estelafs:
# Description
Created the shared component in the UI library FixedWrapperComponent with a floating div at the bottom of the page where we can inject the context (action buttons etc), and used it in the following pages: 
- application settings
- API configurations
- reference data
- form page
- role channels
- role details
- user details
- profile page

## Useful links
- Please insert link to ticket: https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/75138

## Type of change

# How Has This Been Tested?
Go to the Reference Data, choose one to edit and add enough data to need the scroll bar to see the div floats above the page always
Or in when adding a new record or updating (without using the modal), go to a survey when a page has enough questions to need the scroll 

## Screenshots
Reference data:
[refdata.webm](https://github.com/ReliefApplications/oort-frontend/assets/28535394/3fdc634c-9a5d-42d0-bd32-7c9f4d04ea02)

Form:
[form.webm](https://github.com/ReliefApplications/oort-frontend/assets/28535394/558e2710-55a8-4838-9d03-37c8f8ebda20)

API configuration:
[api-config.webm](https://github.com/ReliefApplications/oort-frontend/assets/28535394/84394f5e-3fe2-4964-aab0-9f014231ef54)

Profile:
[profile.webm](https://github.com/ReliefApplications/oort-frontend/assets/28535394/6b2f1995-a569-42a2-8ee8-2dd84e7538c6)

# Checklist:

( * == Mandatory ) 


AntoineRelief:
@estelafs 
it's not just about the page you edited, it's supposed to be a general update of the system
btw, that should certainly be put in a shared component ( not the buttons, but the fixed wrapper, where you could have for example a <ng-content> in it, so you can inject the buttons, depending on the view )

In addition, add a shadow on top, when the user is not at very bottom of the page ( so we can see that the block is floating )

estelafs:
@AntoineRelief I had two problems that i couldn't solve, and after beat my brains out i think it's time for someone else to take a look:

- I had a problem with the `position: fixed` CSS rule and the `transform: translateX(0%)` used for the animations in the ui-tabs: the translateX breaks the fixed (like in the screenshots), but if we remove the translateX the div is fixed and above everything as expected. I couldn't find a way to fix it or replaced it. Pages that used the ui-tabs and need the new floating fixed div where it happens: `safe-user-details`, `safe-role-channels` and `safe-role-details`:
Roles:
[roles.webm](https://github.com/ReliefApplications/oort-frontend/assets/28535394/8c19a81e-a5dd-441e-b95d-1ec4255d044e)
User details:
users

- And i couldn't find a way to add the shadow on top of the div only "when the user is not at the very bottom of the page",  i didn't find where to listen to the scroll event that worked for all components so we could identify when the scroll hits the bottom (neither with addEventListener, renderer2.listen, `@HostListener`)

AntoineRelief:
> * And i couldn't find a way to add the shadow on top of the div only "when the user is not at the very bottom of the page",  i didn't find where to listen to the scroll event that worked for all components so we could identify when the scroll hits the bottom (neither with addEventListener, renderer2.listen, `@HostListener`)

For this part, css should be able to do it by itself I think ( like using :before / :after )

AntoineRelief:
One issue I see with this one

Capture d’écran 2023-09-19 à 16 48 02

We may have some troubles as the parent fixed is not calculated on the correct element
if there is any way to make sure the width is calculated based on the page content ( see the green / blue below ):

Capture d’écran 2023-09-19 à 16 49 48

unai-reliefapp:
@AntoineRelief @estelafs There is not an easy fix for this one.
Position fixed always work against the document body, except if nearest parent contains a transform: translate property, which would take as reference.
The problem is that we are using the sidenav as the main layout, and contains a default padding, so even if we try to keep the fix position and adapt to the current width, got this:
image

Other problem is that with the ui-tabs, and overall the whole sidenav content, elements are not taking all the available height, which leads to have this fixed button(even relative to the content, not the body) to be placed underneath the content but in different places as the content layout does not have a default height, it just expands related to it's own content.

In order to integrate this floating wrapper we would have to make major changes in the main structure if we want to keep it as the requirements ask.

But it's going to affect the whole application, which transforms this ticket into a huge task.
Other solution is to keep this PR, knowing that the fixed wrapper would take the whole body width and rearrange the buttons in the ui-tabs to take them outside the parent so the transform property does not affect it and arrange the logic to view child the component instead of using events.

But we'll have to take care of the main structure to have a clean and easy one to handle, because tomorrow who knows what can it be
Moving the floating outside the ui-tabs
image

But the width is relative to the body document:
image

KR, Unai

AntoineRelief:
@unai-reliefapp 
What about other position, like sticky?

unai-reliefapp:
> @unai-reliefapp What about other position, like sticky?

@AntoineRelief @estelafs PR ready to check 👍

libs/ui/src/lib/sidenav/layout/layout.service.ts
```diff
@@ -1,4 +1,4 @@
-import { Injectable } from '@angular/core';
+import { Injectable, TemplateRef } from '@angular/core';
```
AntoineRelief, line 1: @unai-reliefapp 
why moving the service there?

unai-reliefapp: you can check the last commit message, to avoid circular dependencies now that the service is used in UI and SAFE libraries.
as it's sidenav related almost entirely i moved it inside the sidenav folder

AntoineRelief:
@unai-reliefapp 
In some cases, the content below the bottom element is not visible, and not accessible by scrolling

Capture d’écran 2023-09-20 à 16 31 27

AntoineRelief:
@unai-reliefapp 
In some cases, the content below is over the bottom element:
Capture d’écran 2023-09-20 à 16 28 17

PR #1897, opened 2023-09-27
feat/AB#73404_enable_column_record_specific_dashboard_in_grid

RenzoPrats:
# Description

Added option in the settings of grid and summarycards to have the possibility to display a column that onclick redirect to the record-specific dashboard of certain page. Also allowed to go back to the previous grid view from the redirected view page.
 
## Useful links

[ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/73404)

## Type of change

# How Has This Been Tested?

It has been tested setting a dashboard to have a redirect record-specific dashboard column and testing if it's working correctly

## Screenshots

Peek 29-09-2023 10-23

# Checklist:

( * == Mandatory ) 


estelafs (approved):
- Update some comments to keep consistency and added a tooltip to the record id checkbox to inform that is used to redirect the user to the context view of the record.

- Fixed page url generation because it wasn't working in the front-office (the 'applications' url params were always added, since the url need used was saved always in the back-office)

Taking this into consideration, everything is working as expected.

libs/shared/src/lib/components/ui/core-grid/grid/grid.component.ts
```diff
@@ -240,6 +252,22 @@ export class GridComponent
...
+    if (state.skip && state.take) {
+      const page: PageChangeEvent = {
+        skip: state.skip,
+        take: state.take,
+      };
+      // clear the state
+      this.location.replaceState(this.location.path(), undefined, {
+        navigationId: state.navigationId,
+      });
+      setTimeout(() => {
```
unai-reliefapp: Hi @RenzoPrats , every time we use a timeout or interval we should now store the result and clean it up as a regular observable in order to avoid memory leaks

Declare a new property like
`private pageChangeTimeoutListener!: NodeJs.Timeout`
```
// Add this before any assignation to timeout listeners to clear any previously stored value
// And also add this check in the components ngOnDestroy hook to clean it up definitely on component destruction
if(this.pageChangeTimeoutListener){
  clearTimeout(this.pageChangeTimeoutListener);
}
this.pageChangeTimeoutListener = setTimeout(() => {
```

libs/shared/src/lib/components/widgets/grid-settings/tab-actions/tab-actions.component.ts
```diff
@@ -53,5 +63,68 @@ export class TabActionsComponent {
...
+  ngOnInit(): void {
+    this.showSelectPage = this.formGroup.controls.actions.get(
+      'showRecordDashboard'
+    )?.value;
+    // Add available pages to the list of available keys
+    const application = this.applicationService.application.getValue();
+    this.pages = this.getPages(application);
+    this.formGroup.controls.actions
+      .get('showRecordDashboard')
+      ?.valueChanges.subscribe((val: boolean) => {
```
unai-reliefapp:
```suggestion
      ?.valueChanges.pipe(takeUntil(this.destroy$).subscribe((val: boolean) => {
```

unai-reliefapp:
@AntoineRelief Would not be better to have the checkbox to set the record id inline with the menu select instead of in block and with so much space in between?
Currently
image

Proposal:
image

Or at least without that many space in between select and checkbox:
image

Also just to confirm, the pages available should be the ones in the same application of the current widget settings that we are editing?

KR, Unai

unai-reliefapp (changes requested):
Is there a reason why for the actions in the layout and aggregations settings are seen this way in this branch 😕 
image

AntoineRelief:
@unai-reliefapp 
what I'd suggest is to keep that in two lines, but to have some space on the left or any indicator that would indicate that these two options are linked to the toggle "show record dashboard view"
Btw, I've already said that this name: "show record dashboard view" is not correct, and I'd prefer something like:
"navigate to page"

AntoineRelief:
> Is there a reason why for the actions in the layout and aggregations settings are seen this way in this branch 😕 image

this is fixed in beta @unai-reliefapp

unai-reliefapp:
> @unai-reliefapp what I'd suggest is to keep that in two lines, but to have some space on the left or any indicator that would indicate that these two options are linked to the toggle "show record dashboard view" Btw, I've already said that this name: "show record dashboard view" is not correct, and I'd prefer something like: "navigate to page"

Something like this?
image

Also please @RenzoPrats update the translation with the "Navigate to page" label plus also add the missing values for the french language

RenzoPrats:
> > @unai-reliefapp what I'd suggest is to keep that in two lines, but to have some space on the left or any indicator that would indicate that these two options are linked to the toggle "show record dashboard view" Btw, I've already said that this name: "show record dashboard view" is not correct, and I'd prefer something like: "navigate to page"
> 
> Something like this? image
> 
> Also please @RenzoPrats update the translation with the "Navigate to page" label plus also add the missing values for the french language

Hi @unai-reliefapp , the repairs were made, please check it again, thanks!

libs/shared/src/lib/components/widgets/grid-settings/grid-settings.forms.ts
```diff
@@ -139,6 +139,11 @@ export const createGridWidgetFormGroup = (id: string, configuration: any) => {
       addRecord: [get(configuration, 'actions.addRecord', false)],
       export: [get(configuration, 'actions.export', true)],
       showDetails: [get(configuration, 'actions.showDetails', true)],
+      showRecordDashboard: [
```
AntoineRelief: @RenzoPrats 
cc @unai-reliefapp 
When I asked to change the text, I also meant to change that
I'll do it, but the name of the settings like that do not make sense based on new display

libs/shared/src/lib/components/widgets/grid-settings/tab-actions/tab-actions.component.html
```diff
@@ -3,7 +3,7 @@
     <h2>
       {{ 'components.widget.settings.grid.actions.title' | translate }}
     </h2>
-    <div class="flex flex-col gap-2">
+    <div class="flex flex-col gap-2 mb-2">
```
AntoineRelief: @RenzoPrats 
cc @unai-reliefapp 
usually, you don't want the item which would always be displayed to set the margin
the margin should be set by the elements that appear based on conditions

libs/shared/src/lib/components/widgets/grid-settings/grid-settings.forms.ts
```diff
@@ -139,6 +139,11 @@ export const createGridWidgetFormGroup = (id: string, configuration: any) => {
...
+        get(configuration, 'actions.showRecordDashboard', false),
+      ],
+      pageIdUrl: [get(configuration, 'actions.pageIdUrl', '')],
```
AntoineRelief: @RenzoPrats cc @unai-reliefapp 
I don't like the name of this parameter
pageIdUrl, what is it? an id or an url?
I think it's an id, so just use id

libs/shared/src/lib/components/widgets/grid-settings/tab-actions/tab-actions.component.html
AntoineRelief: @RenzoPrats cc @unai-reliefapp 
looking at it again, I understand it may be because the overflow hides the animation of the toggle
in that case, you can put the margin or padding at parent level ( around all the toggles )

libs/shared/src/lib/components/widgets/grid-settings/tab-actions/tab-actions.component.html
```diff
@@ -17,6 +17,36 @@ <h2>
...
+            [placeholder]="'components.widget.settings.grid.actions.pageId.placeholder' | translate"
+          >
+            <ui-select-option *ngFor="let page of pages" [value]="page.urlParams">
+              {{ page.name }}
+            </ui-select-option>
+          </ui-select-menu>
+        </div>
+        <div>
+          <ui-checkbox formControlName="useRecordId">
+            <ng-container ngProjectAs="label">
```
AntoineRelief: @RenzoPrats cc @unai-reliefapp 
not enough space between both

libs/shared/src/lib/components/widgets/grid-settings/tab-actions/tab-actions.component.html
unai-reliefapp: true, plus I don't see any left padding according to the toggle label as I placed in the image
Please also add that @RenzoPrats , thanks

libs/shared/src/lib/components/widgets/summary-card-settings/summary-card-settings.component.ts
```diff
@@ -74,6 +77,21 @@ const createSummaryCardForm = (def: any) => {
...
+    actions: fb.group({
+      delete: [get(settings, 'actions.delete', true)],
+      history: [get(settings, 'actions.history', true)],
+      convert: [get(settings, 'actions.convert', true)],
+      update: [get(settings, 'actions.update', true)],
+      inlineEdition: [get(settings, 'actions.inlineEdition', true)],
+      addRecord: [get(settings, 'actions.addRecord', false)],
+      export: [get(settings, 'actions.export', true)],
+      showDetails: [get(settings, 'actions.showDetails', true)],
+      showRecordDashboard: [
```
AntoineRelief: @RenzoPrats 
cc @unai-reliefapp 
perhaps we could have pageId and useRecordId as sub-parameters of this one, otherwise, all the properties are at the same level, which does not really make sense for pageId and useRecordId, as they would only appear if the navigation to page is enabled

libs/shared/src/lib/components/widgets/grid-settings/grid-settings.forms.ts
RenzoPrats: It's the url of the page, so I changed the name to pageUrl and also the text displayed to "Page source url".

RenzoPrats:
Changes requested were made, I also did some fix in the summary card when it's displayed as grid to it work correctly. Please review it again @AntoineRelief @unai-reliefapp and let me know if some issue stil.

unai-reliefapp:
Does this column title suit you @AntoineRelief ?
image

Also in the page dropdown we are also displaying form type and workflow type pages of the current application.
Should we display a list of pages of all applications, just the pages of the current application, and if so, only the dashboard types?
image

Besides that I think it's ready.
I refactored the grid actions form group to a new function for both grids and summary cards, plus also I change the page change through location in order to happen once the data is ready, not using a timeout. Also fixed the spacing in order to be align to the rest of labels

KR, Unai

libs/shared/src/lib/components/ui/core-grid/core-grid.component.ts
```diff
@@ -869,6 +905,22 @@ export class CoreGridComponent
...
         break;
       }
+      case 'recordDashboard': {
+        if (event.item) {
+          let fullUrl = this.getPageUrl(event.pageUrl as string);
+          if (event.useRecordId) {
+            const recordId = event.item.id;
+            fullUrl = `${fullUrl}?id=${recordId}`;
+          }
+          this.location.replaceState(this.location.path(), undefined, {
```
AntoineRelief: @RenzoPrats 
@unai-reliefapp 

could one of you explain me why we need that?

unai-reliefapp: @AntoineRelief  the location thing?

AntoineRelief: yes, why do we have that ? I don't see any comment about this, and it's not used anywhere else in the code, so I don't get it

unai-reliefapp: WHO wanted to save the current pagination state of the table from where the record was open in order to load the same paginated list once the user goes back from the open record in the dashboard view.
The location would save the that state so when the user goes back from the record and loads the application again, the grid would trigger a pagination after the data is loaded using the previously saved location state, which would be cleaned up after it's used.
I left a quick explanation also in my last comment in the PR.
@RenzoPrats please add comments in this part of the code why and how the location is used for loading previous paginated list in the grid.

AntoineRelief:
moved the code to beta branch
I need someone to remove the feature discussed above about being able to save the location for pagination
it's too soon for this feature, as we would need additional things about filtering / sorting to be saved as well

libs/shared/src/lib/components/ui/core-grid/core-grid.component.ts
```diff
@@ -309,9 +329,12 @@ export class CoreGridComponent
     private translate: TranslateService,
     private dateTranslate: DateTranslateService,
     private applicationService: ApplicationService,
-    private contextService: ContextService
+    private contextService: ContextService,
+    private router: Router,
+    private location: Location
```
AntoineRelief: to be removed

libs/shared/src/lib/components/ui/core-grid/core-grid.component.ts
AntoineRelief: to be removed

libs/shared/src/lib/components/ui/core-grid/grid/grid.component.ts
```diff
@@ -65,11 +72,16 @@ const matches = (el: any, selector: any) =>
   selector: 'shared-grid',
   templateUrl: './grid.component.html',
   styleUrls: ['./grid.component.scss'],
-  providers: [PopupService, ResizeBatchService],
+  providers: [
+    PopupService,
+    ResizeBatchService,
+    Location,
```
AntoineRelief: to be remvoed

unai-reliefapp:
Would you mind to remove those lines for the Location @RenzoPrats ?
Thank you

RenzoPrats:
> Would you mind to remove those lines for the Location @RenzoPrats ? Thank you

Hi @unai-reliefapp, No that's fine, I will remove it.

AntoineRelief:
@unai-reliefapp 
just fixed the issue with validators on main branch, thanks for pointing out the issue 👍

PR #1928, opened 2023-10-04
Fix missing translations

MwanPygmay:
# Description

Fixed missing translations


libs/shared/src/lib/components/role-summary/role-features/role-forms/role-forms.component.html
```diff
@@ -1,5 +1,5 @@
 <div *ngIf="filteredPages.length">
-  <h2>Forms</h2>
+  <h2>{{ 'common.form' | translate }}</h2>
```
AntoineRelief: @MwanPygmay 
you forgot that common.form is an object:

Capture d’écran 2023-10-05 à 10 20 21

PR #1930, opened 2023-10-04
Feat/ab#69019 allow user with read permission duplicate applications and page

RenzoPrats:
# Description

Allowed users with read permission duplicate pages and applications instead of only with update permissions

## Useful links

[ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/69019)
[backend](https://github.com/ReliefApplications/oort-backend/pull/763)

## Type of change

# How Has This Been Tested?

It has been tested verifying if users with read permission can now duplicate pages and applications.

# Checklist:

( * == Mandatory ) 


unai-reliefapp (changes requested):
Hi @RenzoPrats , please double check and confirm that all duplicate actions in the page are updated accordingly
image
KR, Unai

RenzoPrats:
> Hi @RenzoPrats , please double check and confirm that all duplicate actions in the page are updated accordingly image KR, Unai

Hi @unai-reliefapp, I think the only duplicate action that was not updated was in the settings of the aplication because I'm not sure if an user without canUpdate permission can acess it, but now I updated it also. Thank you!

AntoineRelief:
merged newest version of beta in it
things I see missing:
- with the new way to display settings of applications, I don't believe we cover a case where you can read the application but not update it. In that case, we must check that:
  - only people with read / update access of the app can open settings of a page
  - if you just have read access to the app, you should only see some part of the settings ( or see them as deactivated, that's certainly better )
 
Capture d’écran 2023-10-16 à 13 03 35

Capture d’écran 2023-10-16 à 13 03 24

so from screenshots above, only the "duplicate" action should be accessible to an user that can only read the app
you need to use: @casl/angular in order to do that

apps/back-office/src/app/dashboard/pages/dashboard/dashboard.component.html
```diff
@@ -29,6 +29,7 @@ <h1 class="!m-0 overflow-hidden text-ellipsis">
           <!-- Settings -->
           <ui-button
             *ngIf="!formActive"
+            [disabled]="!('read' | able : 'Application')"
```
AntoineRelief: @RenzoPrats 
it's not the logic to apply
there, the user can see the page, and, perhaps, can edit it
so in all cases, it will be able to see the settings
no need to test anything, and the test there is not correct

apps/back-office/src/app/dashboard/pages/applications/applications.component.html
```diff
@@ -107,7 +107,7 @@ <h2 class="m-0">{{ 'pages.applications.all' | translate }}</h2>
               <button
                 uiMenuItem
                 (click)="onClone(element)"
-                *ngIf="element.canDelete"
+                *ngIf="element.canSee"
```
AntoineRelief: @RenzoPrats 
it's a bit confuse
one elements where you should have put the test on ability to create a new application, you put tests on the element itself, but on some others places, where the logic should have been thought on the element, you put a test about the global model 
i'll create a documentation about that, I see it's still confuse

PR #1933, opened 2023-10-05
AB#76508_fix_grids_loading_multiple_times

MwanPygmay:
# Description

Fix the charts loading multiple times. Not 100% satisfied with my fix, as it involves a big throttle time that freezes context filters changing. If someone has a better fix for this, please share.

## Useful links

[Please insert link to ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/76508/)

## Type of change

# How Has This Been Tested?

Added chart, tried to reload page and activating/deactivating dashboard filter

## Screenshots

# Checklist:

( * == Mandatory ) 

### DIFF
--- a/apps/back-office/src/app/dashboard/pages/dashboard/dashboard.component.ts
+++ b/apps/back-office/src/app/dashboard/pages/dashboard/dashboard.component.ts
@@ -331,7 +331,9 @@ export class DashboardComponent
             : '';
           this.buttonActions = this.dashboard.buttons || [];
           this.showFilter = this.dashboard.showFilter ?? false;
+          this.contextService.isFilterEnabled.next(this.showFilter);
         } else {
+          this.contextService.isFilterEnabled.next(false);
           this.snackBar.openSnackBar(
             this.translate.instant('common.notifications.accessNotProvided', {
               type: this.translate
@@ -687,6 +689,7 @@ export class DashboardComponent
             this.handleDashboardMutationResponse(errors, data);
           },
           complete: () => {
+            this.contextService.isFilterEnabled.next(this.showFilter);
             this.loading = false;
           },
         });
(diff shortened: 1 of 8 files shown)


unai-reliefapp:
Ey @MwanPygmay , I pushed a new approach.

What I did is merge the isFilterEnabled and filter value into the same stream(as it's used in all widget same way to trigger filter changes) and only trigger them whenever a filter value changes, on enabling filter we are adding the dashboard filter component which would emit a default value on init and on destroy, therefor would be always correctly triggered but we only would take care of filter values that are different.

I also fix some timeout listeners to avoid memory leaks 👍

AntoineRelief:
@unai-reliefapp 
I don't see the same logic applied to the grid widget, any reason why?

AntoineRelief:
@unai-reliefapp 
if you access a page with a dashboard filter, the value of the isFilterEnabled seems to be false, and so, the context filter cannot be passed to the widgets that would use it

putting back the ticket into todo

AntoineRelief (changes requested):
not working as it is

the dashboard filter isn't passed to the widgets if you open a page that uses it

AntoineRelief:
video:
https://www.loom.com/share/1c3c387dcddf4210aa4b7ebca7546d81?sid=b89a9ee7-e298-474d-8cb9-9eb5b1cae455

unai-reliefapp:
> @unai-reliefapp I don't see the same logic applied to the grid widget, any reason why?

the only grid using that logic was the aggregation-grid and keep it there only.
Core-grid component is subscribing but just to the filter$ value, even though that filter value is not used in the reloadData method: https://github.com/ReliefApplications/oort-frontend/blob/AB%2376508_fix_grids_loading_multiple_times/libs/shared/src/lib/components/ui/core-grid/core-grid.component.ts#L318

unai-reliefapp:
@AntoineRelief Last commit makes the related widget to only listen to the filter values, as the isFilterEnable property it's only used in the dashboard filter injection method.
So we still keep sending the isFilterEnable value on toggle and application load, but the widget only listen to the filter value changes and the dashboard filter injection method will handle the apply or not of filters if enable or not.
On enable/disable dashboard filter this will still work as on init and on destroy we are triggering a filter value(on destroy an empty object, no filter), so with this the whole flow is closed and reliable
Thanks fo the spot
KR, Unai

PR #1938, opened 2023-10-06
Oort feat/HIT-98 add option to request confirm when auto modifying fields in quick action

GuilhermeGabriel:
# Description

Inside a grid, we have the option to set up quick action buttons, one of which is called 'auto modify selected rows.' For this one, I added an option to choose whether confirmation is needed when the user clicks the button.

## Useful links

- Please insert link to ticket: https://oortcloud.atlassian.net/jira/software/projects/OORT/boards/5?selectedIssue=HIT-98

## Type of change

# How Has This Been Tested?

## Screenshots

https://github.com/ReliefApplications/oort-frontend/assets/24783896/c457fb7e-56cc-455c-b3be-ac2a70b7e960

# Checklist:

( * == Mandatory ) 

### DIFF
--- a/libs/shared/src/lib/components/widgets/grid-settings/button-config/button-config.component.scss
+++ b/libs/shared/src/lib/components/widgets/grid-settings/button-config/button-config.component.scss
@@ -1,6 +1,6 @@
 .update-row {
   align-items: end;
-  margin-top: 24px;
+  margin-top: 4px;
 }
 
 .radio-row {
@@ -14,4 +14,8 @@
   padding-left: 24px;
   display: flex;
   flex-direction: column;
+}
+
+.margin-top-need-confirmation-checkbox {
+  margin-top: 4px;
 }
--- a/libs/shared/src/lib/components/widgets/grid-settings/button-config/button-config.component.scss
+++ b/libs/shared/src/lib/components/widgets/grid-settings/button-config/button-config.component.scss
@@ -15,7 +15,3 @@
   display: flex;
   flex-direction: column;
 }
-
-.margin-top-need-confirmation-checkbox {
-  margin-top: 4px;
-}
--- a/libs/shared/src/lib/components/widgets/grid-settings/button-config/button-config.component.ts
+++ b/libs/shared/src/lib/components/widgets/grid-settings/button-config/button-config.component.ts
@@ -113,6 +113,20 @@ export class ButtonConfigComponent
     }
 
     this.formGroup
+      ?.get('needConfirmation')
+      ?.valueChanges.pipe(takeUntil(this.destroy$))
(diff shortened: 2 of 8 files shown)


libs/shared/src/lib/components/widgets/grid-settings/button-config/button-config.component.scss
```diff
@@ -15,3 +15,7 @@
   display: flex;
   flex-direction: column;
 }
+
+.margin-top-need-confirmation-checkbox {
+  margin-top: 4px;
+}
```
matheus-relief: Whenever possible, you should you tailwind classes to keep in line with the project style guidelines. In this case it would be `mt-1`

libs/shared/src/lib/components/widgets/grid-settings/button-config/button-config.component.ts
```diff
@@ -112,6 +112,20 @@ export class ButtonConfigComponent
...
+      .subscribe((value) => {
+        if (value) {
+          this.formGroup
+            ?.get('needConfirmation')
+            ?.setValidators(Validators.required);
+        } else {
+          this.formGroup?.get('needConfirmation')?.clearValidators();
+        }
+        this.formGroup?.get('needConfirmation')?.updateValueAndValidity();
+      });
```
matheus-relief: What's the idea here? I removed it and doesn't seem to break anything. From my understanding all this does is make the field required when it's true and not required when it's false, which doesn't make much sense to me, but maybe I'm missing something

matheus-relief:
@GuilhermeGabriel Made a few adjustments besides the ones in the comments above:
* Renamed `needConfirmation` to `requireConfirmation`, just becuase I think that reads better
* Changed the logic in the `grid.component.ts` file to avoid code duplication (check [here](https://github.com/ReliefApplications/ems-frontend/pull/1938/commits/1363c1b65dee49227fa43fdc19e5e8bfbc08c2f3#diff-8cd7eb861fede6d8d41c971209cc92b0caffb928663446510800ce61fa8030d2))

I wont delete this branch for now in case you can explain to me that bit of code I commented out

libs/shared/src/lib/components/widgets/grid-settings/button-config/button-config.component.ts
GuilhermeGabriel: Yes, you're correct, i added this piece of code by mistake. You can delete it.

libs/shared/src/lib/components/widgets/grid-settings/button-config/button-config.component.scss
GuilhermeGabriel: Ok!

PR #1997, opened 2023-10-24
feat/AB#77900_ABC-allow-search-in-select

unai-reliefapp:
# Description

feat: add search bar to menu-select component, plus updated empty state for the search results 
fix: survey component application dropdown ui update on close menu option list using a custom observable to listen to UI changes(dropdown close and search)
fix: trim select option list label

## Useful links

- [Link to ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/77900)

## Type of change

# How Has This Been Tested?

Please refer to video below

## Screenshots

Behavior with the new search bar on select-menu component:

https://github.com/ReliefApplications/ems-frontend/assets/123092672/ca8f45c4-a2de-4753-ab27-4181bb4112f5

# Checklist:

( * == Mandatory ) 


libs/ui/src/lib/select-menu/select-menu.component.ts
```diff
@@ -120,7 +153,9 @@ export class SelectMenuComponent
         if (
           !(
             this.el.nativeElement.contains(event.target) ||
```
AntoineRelief, line 159: @unai-reliefapp 
could you update storybook stories in order to reflect the change and demonstrate it?

AntoineRelief (changes requested):
@unai-reliefapp 
the feature is working fine, but I can see a styling issue in the select:

The search appears to be outside the box of the ul below it

I think you should move most of the classes to the html element that encapsulates the search and the ul

for example, there is a border and rounded effect in the url, and it does not look fine with the search on top of it

Capture d’écran 2023-10-24 à 13 11 01

AntoineRelief:
@unai-reliefapp 
Another required change
When we close the select, the search should be erased

AntoineRelief:
@unai-reliefapp 
it looks quite similar to what we have in the graphQL select ( even if the filter does not do the same thing, as the graphql select would instead send an updated query to the back

perhaps you could just use one logic to display or not the search control in both, and make sure you can override the default behavior of the search in the select, by using the custom graphql select logic instead

unai-reliefapp:
> @unai-reliefapp it looks quite similar to what we have in the graphQL select ( even if the filter does not do the same thing, as the graphql select would instead send an updated query to the back
> 
> perhaps you could just use one logic to display or not the search control in both, and make sure you can override the default behavior of the search in the select, by using the custom graphql select logic instead

I know what do you mean but with that change, the changes needed in the whole application go huge way up as we have to update all the current selectors inside the application, as calling the component graphql select it's not clear as it would handle local search also.

As it's done now, it's just to add a filterable property to the select-menu items and we are good to go, as if we need to use a query it's already done by using the graphql select component.

If this is fine I'll change the logic to the graphql component, sure

KR, Unai

unai-reliefapp:
> @unai-reliefapp Another required change When we close the select, the search should be erased

I dont know if you should also have to update the related ticket with this feature, but just in case

KR, Unai

AntoineRelief:
> > @unai-reliefapp it looks quite similar to what we have in the graphQL select ( even if the filter does not do the same thing, as the graphql select would instead send an updated query to the back
> > perhaps you could just use one logic to display or not the search control in both, and make sure you can override the default behavior of the search in the select, by using the custom graphql select logic instead
> 
> I know what do you mean but with that change, the changes needed in the whole application go huge way up as we have to update all the current selectors inside the application, as calling the component graphql select it's not clear as it would handle local search also.
> 
> As it's done now, it's just to add a filterable property to the select-menu items and we are good to go, as if we need to use a query it's already done by using the graphql select component.
> 
> If this is fine I'll change the logic to the graphql component, sure
> 
> KR, Unai

@unai-reliefapp 
let's put that in another ticket
true that it's a lot of changes

unai-reliefapp:
@AntoineRelief added requested changes in the last commit 👍 

https://github.com/ReliefApplications/ems-frontend/assets/123092672/5c019756-1eee-4967-8eae-c549c1fcf9df

PR #2029, opened 2023-10-31
Fix/ab#78168 allow other fields be passed as query parameter

RenzoPrats:
# Description

Allowed in navigate to page settings to select another field than id to be passed as query parameter. If this field is a reference data question, the valueField is taken and passed as parameter of the field.

## Useful links

[ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/78168)

## Type of change

# How Has This Been Tested?

It has been tested creating a context dashboard with some reference data, and setting up the grid action navigate to page to navigate to this dashboard using as id the question field that has this reference data.

## Screenshots

Peek 09-11-2023 08-42

# Checklist:

( * == Mandatory ) 


libs/shared/src/lib/components/ui/core-grid/core-grid.component.ts
```diff
@@ -273,7 +273,7 @@ export class CoreGridComponent
     showDetails: true,
     navigateToPage: false,
     navigateSettings: {
-      useRecordId: false,
+      recordField: '',
```
AntoineRelief: @RenzoPrats 
just use "field"

libs/shared/src/lib/components/ui/core-grid/core-grid.component.ts
```diff
@@ -898,9 +898,29 @@ export class CoreGridComponent
...
         if (event.item) {
           let fullUrl = this.getPageUrl(event.pageUrl as string);
-          if (event.useRecordId) {
-            const recordId = event.item.id;
-            fullUrl = `${fullUrl}?id=${recordId}`;
+          if (event.recordField) {
+            const field = event.recordField;
+            let recordField = '';
+            if (field === 'id') {
+              recordField = event.item.id;
```
AntoineRelief: @RenzoPrats 
I think it's too complex
You can certainly just use lodash to get the field from the item
( using get method )

libs/shared/src/lib/components/widgets/common/tab-actions/tab-actions.component.ts
```diff
@@ -78,12 +84,47 @@ export class TabActionsComponent
...
+    private apollo: Apollo,
+    private referenceDataService: ReferenceDataService
+  ) {
     super();
   }
 
   ngOnInit(): void {
+    if (this.formGroup.get('resource')?.value) {
+      this.apollo
+        .query<ResourceQueryResponse>({
```
AntoineRelief: @RenzoPrats 
don't use this logic there, it's too complex and it doesn't work in all cases
for example, in the grid-settings component, you have the "fields" property
I think you can just use this one because it should contain all the logic you need to select a field from the selected datasource
ONLY available fields should be selectable, not all fields

AntoineRelief:
@RenzoPrats 
some changes to be done
i already added many comments on the PR
Please take that as a priority

overall logic is fine, but there are still elements I don't agree with

In addition to the comments already left on files, you should also have a look at this select:
Capture d’écran 2023-11-08 à 17 44 27

Which is incorrect ( considering that country is a reference data )
ALL fields of the reference data should be selectable
So in that case, you can use something like that:
Capture d’écran 2023-11-08 à 17 47 05 (1)

or like what is defined in the field-dropdown file ( where you can expand the fields )

perhaps you can even use this component in fact

AntoineRelief:
@RenzoPrats 
you can have a look at the code when you have time
I simplified it, so the fields are passed by the parent element ( grid settings or summary card settings )

PR #2059, opened 2023-11-09
feat/AB#78949_possibility_to_inject_aggregation_in_html_widgets

RenzoPrats:
# Description

Feat: Added possibility to inject aggregation result in html widgets.

## Useful links

[ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/78949)

## Type of change

# How Has This Been Tested?

It has been tested verifying if it's possible to select an aggregation and use it in the autocompleter in the editor in html widgets.

## Screenshots

Peek 14-11-2023 09-18

# Checklist:

( * == Mandatory ) 


MwanPygmay (commented):
- Tested, indeed works, but options available for now are very restricted (the aggregation just appears in the propositions, but you cannot get data from it).
- Added mapping for the new aggregation

libs/shared/src/lib/components/widgets/common/aggregation-settings/aggregation-settings.component.ts
```diff
@@ -0,0 +1,235 @@
...
+      query: GET_RESOURCES,
+      variables: {
+        first: ITEMS_PER_PAGE,
+        sortField: 'name',
+      },
+    });
+
+    this.formGroup
+      .get('aggregation.resource')
+      ?.valueChanges.subscribe((val: any) => {
```
unai-reliefapp:
```suggestion
      ?.valueChanges.pipe(takeUntil(this.destroy$)).subscribe((val: any) => {
```

libs/shared/src/lib/components/widgets/common/aggregation-settings/aggregation-settings.component.ts
```diff
@@ -0,0 +1,235 @@
...
+      },
+    });
+    dialogRef.closed.pipe(takeUntil(this.destroy$)).subscribe((value) => {
+      if (value) {
+        if (typeof value === 'string') {
+          this.formGroup.get('aggregation.id')?.setValue(value);
+        } else {
+          this.formGroup.get('aggregation.id')?.setValue((value as any).id);
+          this.formGroup.get('aggregation.name')?.setValue((value as any).name);
+          this.selectedAggregation = value;
```
unai-reliefapp, line 197: If value type set from the modal is string(aggregation id) we still have an aggregation selected in the formgroup, but the `selectedAggregation` property could be undefined, has this been thoroughly checked?(on canceling, saving and opening again modal, on switching tabs, on close and openning settings etc etc)

e.g. `editAggregation` method won't work as it's checking `if (value && this.selectedAggregation)` and if the first added aggregation only sets the id like it's done here, `selectedAggregation` is not updated and even if an aggregation id is set, cannot be edited

KR, Unai

libs/shared/src/lib/components/widgets/editor-settings/editor-settings.component.ts
```diff
@@ -122,6 +142,13 @@ export class EditorSettingsComponent implements OnInit, AfterViewInit {
           }
         });
     }
+
+    // update fields when change aggregation
+    this.widgetFormGroup
+      .get('aggregation')
+      ?.valueChanges.subscribe((val: any) => {
```
unai-reliefapp:
```suggestion
      ?.valueChanges.pipe(takeUntil(this.destroy$)).subscribe((val: any) => {
```

unai-reliefapp (changes requested):

AntoineRelief:
@RenzoPrats @unai-reliefapp @MwanPygmay 
I will need to close this one, as it won't work
my bad for this one, I didn't provided enough information in the ticket

I was expecting to have the possibility to put many aggregations, like that:

{{aggregation(id)[0].count}} for example, which should be transformed when loading the widget
this last part is missing

Also, one aggregation per widget is not enough, and it should be possible to use more than one
In addition, I'll remove the possibility to do it in summary cards, as I think it will generate conflicts

PR #2087, opened 2023-11-21
Fix/ab#79454 wrong calculation grid columns

LenaelleL:
# Description

Includes the width of non-field columns in the calculation of column width for grid widgets

## Useful links

- Please insert link to ticket : https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/79454

## Type of change

# How Has This Been Tested?

- [x] Tested with a grid widget an different numbers of columns

## Screenshots

After modifications :
image

# Checklist:

( * == Mandatory ) 

# More explanation


Maximo-reliefapps (commented):
The issue has been fixed, the grid isn't showing the overflow anymore when you only have one column, all good for me.

libs/shared/src/lib/components/ui/core-grid/grid/grid.component.ts
```diff
@@ -979,7 +979,8 @@ export class GridComponent
    */
   private setColumnsWidth() {
     const gridElement = this.gridRef.nativeElement;
-    const gridTotalWidth = gridElement.offsetWidth;
+    //Subtract the width of non-fields columns (details, actions etc.)
+    const gridTotalWidth = gridElement.offsetWidth - 164;
```
AntoineRelief: @LenaelleRelief @Maximo-reliefapps 
Incorrect logic I fear
the 164 can be replaced by the opposite of this one:

```
const availableColumns = this.columns.filter(
      (column) => !column.hidden && !!column.title && !column.sticky
   );
```
Where we filter out the sticky & no title columns
if you take the rest of the columns, then you have the sticky ones, and you can substract from the size of the widget the total width of such columns ( just using the width parameterof each )

AntoineRelief: I'm saying incorrect as in some cases, the total width of such columns may not be the one you indicate 
Indeed, if you don't allow any action, it's one less or two less columns, so the width can't be 164px

LenaelleL: @AntoineRelief Thanks for the feedback, Renzo will work on it !

libs/shared/src/lib/components/ui/core-grid/grid/grid.component.ts
```diff
@@ -993,6 +991,20 @@ export class GridComponent
...
+      (column) => !column.hidden && !!column.sticky
+    );
+    let totalWidthSticky = 0;
+    stickyFields.forEach((val: any) => {
+      if (val.width) {
+        totalWidthSticky += val.width;
+      }
+    });
+    // fixed amount required
+    totalWidthSticky += 55;
```
AntoineRelief: @RenzoPrats 
where is that coming from? there is no explanation

AntoineRelief: @RenzoPrats 
see video:
https://www.loom.com/share/f1fde6e5f1b243d498f6071c4671d928?sid=21c36304-8e49-4b5b-a0f9-86e0f66af000
I added a few scenarii that are not working as expected

RenzoPrats: @AntoineRelief  this 55 is because the select column of the grid, with the current approach when we get the columns of this grid we do not consider the select column, so we have to add this.

MwanPygmay (approved):
- Tested, indeed removes scrollbar.
- Using hardcoded values does not seem ideal, but I do not have a better solution

libs/shared/src/lib/components/ui/core-grid/grid/grid.component.ts
```diff
@@ -979,24 +979,44 @@ export class GridComponent
...
+    stickyFields.forEach((val: any) => {
+      if (val.width) {
+        totalWidthSticky += val.width;
+      }
     });
+    // fixed amount required for select column
+    if (this.selectable) {
+      totalWidthSticky += 35;
+    }
+    //Subtract the width of non-fields columns (details, actions etc.) and small calculation errors
```
AntoineRelief: please put space between "//" and the text

PR #2122, opened 2023-12-04
bugfix/AB#80650_ABC-reset-default-layout-button-not-appearing-in-fron…

unai-reliefapp:
# Description

refactor: move the admin check of the user to the root grid component using the same getter from the authService from the parent components loading it. Check would be trigger each time the root grid is loaded
refactor: remove not needed dependencies from the refactor aforementioned above
feat: now the admin check is module agnostic for the root grid component, related features are displayed for both back and front office

## Useful links

- [Link to ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/80650)

## Type of change

# How Has This Been Tested?

Please refer to video below

## Screenshots

https://github.com/ReliefApplications/ems-frontend/assets/123092672/3af3ebec-3461-43e6-9d38-35b1bed1cb0f

# Checklist:

( * == Mandatory ) 

# More explanation


RenzoPrats (approved):
Working fine, looks good to me!

AntoineRelief:
@RenzoPrats @unai-reliefapp 
Putting this one back as draft
overall changes are correct, but I think some points were missed:
- for aggregation grid, the button just does nothing, so better hide it in fact
- for classic grid, I don't want the feature to only be accessible to admin. It should be available to all users, so the isAdmin check should be replaced by another check on the fact that the grid uses aggregation or not
- adding to the scope that the title property of the button should be translated

unai-reliefapp:
> @RenzoPrats @unai-reliefapp Putting this one back as draft overall changes are correct, but I think some points were missed:
> 
> * for aggregation grid, the button just does nothing, so better hide it in fact
> * for classic grid, I don't want the feature to only be accessible to admin. It should be available to all users, so the isAdmin check should be replaced by another check on the fact that the grid uses aggregation or not
> * adding to the scope that the title property of the button should be translated

* Display reset layout button only when default table is loaded, as how and where the aggregation table is loaded in the application, with a single input property set in the aggregation table would do the trick
* Free access to reset layout action by not taking in account admin status or any status of the user
* Add translation for reset layout button title

Setting PR back to ready

PR #2129, opened 2023-12-05
Fix/ab#55095 choices by url should be erased automatically if set choices from ref data

GuilhermeGabriel:
# Description

Choices by url properties should be erased automatically if we set choices from reference data.

Priority order for choices:

First choices from ref data,
If not available choices from url,
If not available choices from static choices

## Useful links

- Please insert link to ticket: https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/55095

## Type of change

# How Has This Been Tested?

Adding choices by url, after that setting choices from ref data and seeing if it erase the choices by url automatically.

## Screenshots

https://github.com/ReliefApplications/ems-frontend/assets/24783896/e7be52b8-06c6-484d-badf-79fe44da3158

# Checklist:

( * == Mandatory ) 

# More explanation

### DIFF
--- a/libs/shared/src/lib/survey/global-properties/reference-data.ts
+++ b/libs/shared/src/lib/survey/global-properties/reference-data.ts
@@ -1,5 +1,6 @@
 import { QuestionSelectBase, Question } from '../types';
 import {
+  ChoicesRestfull,
   ItemValue,
   JsonMetadata,
   Serializer,
@@ -36,6 +37,7 @@ export const init = (referenceDataService: ReferenceDataService): void => {
       type: CustomPropertyGridComponentTypes.referenceDataDropdown,
       visibleIndex: 1,
       onSetValue: (obj: QuestionSelectBase, value: string) => {
+        obj.setPropertyValue('choicesByUrl', new ChoicesRestfull());
         obj.choicesByUrl.setData([]);
         obj.setPropertyValue('referenceData', value);
       },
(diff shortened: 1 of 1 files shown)


libs/shared/src/lib/survey/global-properties/reference-data.ts
```diff
@@ -35,6 +35,10 @@ export const init = (referenceDataService: ReferenceDataService): void => {
       category: 'Choices from Reference data',
       type: CustomPropertyGridComponentTypes.referenceDataDropdown,
       visibleIndex: 1,
+      onSetValue: (obj: QuestionSelectBase, value: string) => {
+        obj.choicesByUrl.setData([]);
```
AntoineRelief, line 41: @GuilhermeGabriel 
it seems to be working, but I don't get it visible in the editor of the question
if you have a look at the json editor, indeed the values are removed, but not in the settings
it requires to close the settings of the qu
estion to see the value refreshed
perhaps you can find out if anything can be done about that

For example, in the screenshot above, I just set the reference data value, but the choices by url are still visible, the values are not empty ( but if you save the form, they would be removed, so it's an ui issue )
Capture d’écran 2023-12-05 à 14 20 21

AntoineRelief (changes requested):
@GuilhermeGabriel 
based on comment, putting back as draft

libs/shared/src/lib/survey/global-properties/reference-data.ts
GuilhermeGabriel: Ok!
I added a new commit with the fix (I think this fixes).

https://github.com/ReliefApplications/ems-frontend/assets/24783896/6c1a6569-a173-4033-bf54-ba5d5ed8086e

PR #2154, opened 2023-12-11
fix/AB#81441-calc-methods-in-summary-cards-and-text-widgets

estelafs:
# Description
When a {{calc. method have an argument for a layout, using the {{data. field value, and the layout has a custom style, the calc method will break because of the span with the style element in the HTML. That's because the  applyOperations method that detects the calc method can't get the argument inside the span element.

I updated the applyOperations method to remove all span elements with its styles before check with the regex if any calc method is available int he HTML, and after the calc method is calculated the style and span is put back in the HTML.

My regex spanRegex used to identify and remove the span elements could be improved to only remove span elements inside {{calc. methods, because at the moment all spans are removed and in places where we add in the editor {{data.field}} like this, without calc methods, the style is permanently removed (see screenshots for example), and I couldn't find how to avoid it.

## Useful links
- Please insert link to ticket: https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/81441

## Type of change

# How Has This Been Tested?
Create a summary card or text widget with a layout that has style, then using the {{calc. methods correctly (with numbers fields for example), display the values and results in the cards/text widget
 
## Screenshots
Text widget:
text

Summary card:
summarycard

# Checklist:

( * == Mandatory ) 

# More explanation


libs/shared/src/lib/utils/parser/utils.ts
```diff
@@ -450,8 +449,22 @@ const applyOperations = (html: string): string => {
     `${CALC_PREFIX}(\\w+)\\(([^\\)]+)\\)${PLACEHOLDER_SUFFIX}`,
     'gm'
   );
-  let parsedHtml = html;
-  let result = regex.exec(html);
+
+  // To identify and remove span with style elements to avoid breaking calc functions
+  const spanRegex = /<span[^>]*>([^<]+)<\/span>/g;
```
TaiKamilla: This works fine with normalized strings but the regex is not very accurate with "<" characters
Screenshot 2023-12-12 at 12 12 52

TaiKamilla: try maybe `const spanRegex = /<span[^>]*>([\s\S]*?)<\/span>/g;` it will get nested tags too
Screenshot 2023-12-12 at 12 36 59

TaiKamilla (changes requested):

TaiKamilla (approved):

PR #2160, opened 2023-12-12
feat/AB#81521_ABC-trigger-filter-from-text-and-summary-card

unai-reliefapp:
# Description

feat: listen to filter events from summary card and text widget using custom attributes in the content set from the tinymce
feat: trigger event on click in child elements of the element containing the filter dataset by bubling up through parent nodes
feat: add renderer2 service in order safely handle the loop aforementioned

## Useful links

- [Link to ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/81521)

## Type of change

# How Has This Been Tested?

Please refer to video below

## Screenshots

https://github.com/ReliefApplications/ems-frontend/assets/123092672/63168187-e3ca-49dd-82fe-2046f24a46a8

# Checklist:

( * == Mandatory ) 

# More explanation

### DIFF
--- a/libs/shared/src/lib/components/widgets/editor/editor.component.ts
+++ b/libs/shared/src/lib/components/widgets/editor/editor.component.ts
@@ -171,7 +171,19 @@ export class EditorComponent extends UnsubscribeComponent implements OnInit {
       // Cleanup filter value from the span set by default in the tinymce calculated field if exists
       const cleanContent = filterValue.match(/(?<=>)(.*?)(?=<)/gi);
       const cleanFilterValue = cleanContent ? cleanContent[0] : filterValue;
-      this.contextService.filter.next({ [filterField]: cleanFilterValue });
+      const currentFilters = { ...this.contextService.filter.getValue() };
+      // If current filters contains the field but there is no value set, delete it
+      if (filterField in currentFilters && !cleanFilterValue) {
+        delete currentFilters[filterField];
+      }
+      // Update filter object with existing fields and values
+      const updatedFilters = {
+        ...(currentFilters && { ...currentFilters }),
+        ...(cleanFilterValue && {
+          [filterField]: cleanFilterValue,
+        }),
+      };
+      this.contextService.filter.next(updatedFilters);
     } else {
       const content = this.htmlContentComponent.el.nativeElement;
       const editorTriggers = content.querySelectorAll('.record-editor');
--- a/libs/shared/src/lib/components/widgets/editor/editor.component.ts
+++ b/libs/shared/src/lib/components/widgets/editor/editor.component.ts
@@ -5,6 +5,7 @@ import {
   TemplateRef,
   ViewChild,
   HostListener,
+  Renderer2,
 } from '@angular/core';
 import { SafeHtml } from '@angular/platform-browser';
 import { Apollo } from 'apollo-angular';
@@ -86,6 +87,7 @@ export class EditorComponent extends UnsubscribeComponent implements OnInit {
    * @param gridService Shared grid service
    * @param referenceDataService Shared reference data service
    * @param contextService Context service
+   * @param renderer Angular renderer2 service
    */
   constructor(
     private apollo: Apollo,
@@ -96,7 +98,8 @@ export class EditorComponent extends UnsubscribeComponent implements OnInit {
     private translate: TranslateService,
     private gridService: GridService,
     private referenceDataService: ReferenceDataService,
-    private contextService: ContextService
+    private contextService: ContextService,
+    private renderer: Renderer2
   ) {
     super();
   }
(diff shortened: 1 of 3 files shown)


unai-reliefapp:
Hi @AntoineRelief , this current approach has some limitations:

- As done in this PR, the filter event would only be catch if the user clicks exactly in the element containing the `data-filter ` attribute, not on any child element of the one containing that dataset.
- Do we want to display some kind of tooltip or helper message to the user so they know how they have to set the data-filter in order to make the filter event work? If so, how you want to display that information?
- Currently once a filter is applied from the text-widget or summary-card item, we cannot take the filter back, there is no "clear filter" option for this cases, is that expected?

KR, Unai

AntoineRelief:
> Hi @AntoineRelief , this current approach has some limitations:
> 
> * As done in this PR, the filter event would only be catch if the user clicks exactly in the element containing the `data-filter ` attribute, not on any child element of the one containing that dataset.
> * Do we want to display some kind of tooltip or helper message to the user so they know how they have to set the data-filter in order to make the filter event work? If so, how you want to display that information?
> * Currently once a filter is applied from the text-widget or summary-card item, we cannot take the filter back, there is no "clear filter" option for this cases, is that expected?
> 
> KR, Unai

@unai-reliefapp 

for possible solutions:
- perhaps you can search in parent elements when clicking, up to the one that hosts all the custom html. This way, you would be able to click on child elements, and it's quick to execute
- no tooltip
- it should come as a data-filter-value emitting a null, but it needs to be set up by the admin. as you can see in the design of the ticket, there is one grey box, with "all" written in it. If they click on this one, the filter should be reset

unai-reliefapp:
> > Hi @AntoineRelief , this current approach has some limitations:
> > 
> > * As done in this PR, the filter event would only be catch if the user clicks exactly in the element containing the `data-filter ` attribute, not on any child element of the one containing that dataset.
> > * Do we want to display some kind of tooltip or helper message to the user so they know how they have to set the data-filter in order to make the filter event work? If so, how you want to display that information?
> > * Currently once a filter is applied from the text-widget or summary-card item, we cannot take the filter back, there is no "clear filter" option for this cases, is that expected?
> > 
> > KR, Unai
> 
> @unai-reliefapp
> 
> for possible solutions:
> 
> * perhaps you can search in parent elements when clicking, up to the one that hosts all the custom html. This way, you would be able to click on child elements, and it's quick to execute
> * no tooltip
> * it should come as a data-filter-value emitting a null, but it needs to be set up by the admin. as you can see in the design of the ticket, there is one grey box, with "all" written in it. If they click on this one, the filter should be reset

Okay, I'll update the parent element search 👍

libs/shared/src/lib/components/widgets/editor/editor.component.ts
```diff
@@ -148,13 +154,33 @@ export class EditorComponent extends UnsubscribeComponent implements OnInit {
...
+        filterButtonIsClicked = !!currentNode.dataset.filterField;
       }
-    });
+    }
+    if (filterButtonIsClicked) {
+      const { filterField, filterValue } = currentNode.dataset;
+      // Cleanup filter value from the span set by default in the tinymce calculated field if exists
+      const cleanContent = filterValue.match(/(?<=>)(.*?)(?=<)/gi);
+      const cleanFilterValue = cleanContent ? cleanContent[0] : filterValue;
+      this.contextService.filter.next({ [filterField]: cleanFilterValue });
```
AntoineRelief: @unai-reliefapp 
If I'm correct, there, the value of the filter will be exactly [filterField]: value, which is incorrect, as you're potentially erasing other filter fields

unai-reliefapp: true, my bad, should be

```
const currentFilters = this.contextService.filter.getValue();
const updatedFilters = {
    ...currentFilters,
   [filterField]: cleanFilterValue
};
this.contextService.filter.next(updatedFilters);
```

AntoineRelief: I'll update

unai-reliefapp: Complete code:

```
      const currentFilters = this.contextService.filter.getValue();
      // If current filters contains the field but there is no value set, delete it
      if (filterField in currentFilters && !cleanFilterValue) {
        delete currentFilters[filterField];
      }
      // Update filter object with existing fields and values
      const updatedFilters = {
        ...(currentFilters && { ...currentFilters }),
        ...(cleanFilterValue && { [filterField]: cleanFilterValue }),
      };
      this.contextService.filter.next(updatedFilters);
```

unai-reliefapp: Fixed and pushed @AntoineRelief , problem was I was working with the filter reference instead of a new object, therefor the pairwise check was never passing through and filter was not triggering
```
// Wrong, modifying filter source object
const currentFilters = this.contextService.filter.getValue();

// Correct, modifying filter source object copy
const currentFilters = { ...this.contextService.filter.getValue() };
```


https://github.com/ReliefApplications/ems-frontend/assets/123092672/286cd223-c391-4663-8c38-9fdca0b017fc

PR #2164, opened 2023-12-12
fix/AB#81561_text_widget_editor_autocomplete_not_working_correctly

RenzoPrats:
# Description

Fixed: User has to go to record selection tab to see autocomplete list in text widget editor, for that was created a method to update calc and keys for autocomplete

## Useful links

[ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/81561)

## Type of change

# How Has This Been Tested?

It has been tested opening a text editor widget settings and testing of now the autocomplete is working correctly without the need to go to record selection tab.

## Screenshots

Peek 12-12-2023 19-46

# Checklist:

( * == Mandatory ) 

### DIFF
--- a/libs/shared/src/lib/services/editor/editor.service.ts
+++ b/libs/shared/src/lib/services/editor/editor.service.ts
@@ -14,10 +14,14 @@ import { DOCUMENT } from '@angular/common';
 export class EditorService {
   /** environment variable */
   private environment: any;
-
+  /** edit scroll listener */
   private editorScrollListener!: any;
+  /** active item scroll listener */
   public activeItemScrollListener!: any;
+  /** renderer */
   private renderer!: Renderer2;
+  /** current calc keys for auto completer */
+  private currentKeys: { value: string; text: string }[] = [];
 
   /**
    * Compute the base url based on the environment file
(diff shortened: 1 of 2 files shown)


AntoineRelief:
@RenzoPrats 
closing this one
I don't think it was a correct solution unfortunately

libs/shared/src/lib/services/editor/editor.service.ts
```diff
@@ -14,10 +14,14 @@ import { DOCUMENT } from '@angular/common';
...
   private environment: any;
-
+  /** edit scroll listener */
   private editorScrollListener!: any;
+  /** active item scroll listener */
   public activeItemScrollListener!: any;
+  /** renderer */
   private renderer!: Renderer2;
+  /** current calc keys for auto completer */
+  private currentKeys: { value: string; text: string }[] = [];
```
AntoineRelief, line 24: @RenzoPrats 
I don't like this approach
the keys are so specific to the component calling the service, that it should not be stored in the service
In addition, I think it would work if the service would be per instance, not singleton

AntoineRelief: another approach was to prevent the editor to appear when fields are not yet provided
just doing that, it fix it
I pushed a commit with these changes in beta 👍

RenzoPrats: @AntoineRelief Okay, makes sense, the problem is we cannot define twice the calc and keys and we have to define it when the editor is intiialized so we cannot define a few time later, I thought about this approach but I didn't know if it was correctly to doesn't load the editor instantly so I ending up with this other approach.

PR #2184, opened 2023-12-16
feat/AB#80252_Allow-frontend-to-reload-when-admin-updates/create-a-form

GuilhermeGabriel:
# Description

One issue we usually face while configuring forms/resources is that the system will stop working as expected after saving forms/resources. Indeed, the back-end needs to reload, and the front does not get the changes. We should, if possible, either reload the entire front, either reload the graphql.

## Useful links

- Please insert link to ticket: (I don't have access)

## Type of change

# How Has This Been Tested?

Create a new form, add a field to it, and test if the field appears in the layout editor without reloading the page.

## Screenshots

https://github.com/ReliefApplications/ems-frontend/assets/24783896/6f36a393-a9b1-4896-8007-498a76befe1e

# Checklist:

( * == Mandatory ) 

# More explanation


apps/back-office/src/app/application/pages/add-page/add-page.component.ts
```diff
@@ -202,6 +202,7 @@ export class AddPageComponent extends UnsubscribeComponent implements OnInit {
                 );
 
                 this.onSubmit();
+                window.location.reload();
```
AntoineRelief: @GuilhermeGabriel 
that won't work
you need to wait for the server to be ready again to do that

GuilhermeGabriel: Ok, I tried a new approach!

apps/back-office/src/app/dashboard/pages/form-builder/graphql/queries.ts
```diff
@@ -38,3 +38,121 @@ export const GET_SHORT_FORM_BY_ID = gql`
     }
   }
 `;
+
+/** Graphql request for getting query types */
+export const GET_QUERY_TYPES = gql`
```
AntoineRelief: @GuilhermeGabriel 
I don't like this approach
you're duplicating a query we already see in another service, for a call that will be done only once, so for me it doesn't make sense

you should prefer, for example, build a subject in the query-builder, and, when you detect a change in the subject, you refetch the data

so you can, in the form builder service, just emit an event to the query builder service

matheus-relief: @AntoineRelief do you think a websocket solution would be overkill?

PR #2186, opened 2023-12-18
AB#81834_resources_questions_grid_take_too_much_width'

MwanPygmay:
# Description

Adds a maxwidth property to the grid on creation. Not ideal because resizing the window will cause the grid to be too small/too large, but at least fixes the major use case.

## Useful links

- [Please insert link to ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/81834/)

## Type of change

# How Has This Been Tested?

Reproduce bug like in ticket, grid should now have a reasonable size.

## Screenshots
Bug:
image
Limitations after fix:
fix_limitations

# Checklist:

( * == Mandatory ) 

### DIFF
--- a/libs/shared/src/lib/components/ui/core-grid/grid/grid.component.ts
+++ b/libs/shared/src/lib/components/ui/core-grid/grid/grid.component.ts
@@ -951,27 +951,21 @@ export class GridComponent
       type: string;
       title: string;
     }[] = [];
-    this.fields.forEach((field: any) => {
-      const availableFields = availableColumns.filter((column: any) => {
-        // To correctly size referenceData columns when hiding / show them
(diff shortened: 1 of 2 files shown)


AntoineRelief:
@MwanPygmay 
based on the gif, I think something else can be done
indeed, there is only one column visible in the grid ( instead of the three that you should display ).
It seems the auto calculation of column widths is incorrect there, and instead of dividing the total width so each of the 3 column can take about 1/3 of the available width, there, the first one take all the width
putting back as draft

MwanPygmay:
@AntoineRelief Changed reference data logic, now we unnest the subfields and can correctly get them, width is now correct

PR #2229, opened 2024-01-02
bugfix/AB#82624_tab-widgets-inner-widgets-order-is-not-applied-when-tab-widget-shrinks-to-a-certain-width 

unai-reliefapp:
refactor: some redundant styles and elements for the editor widget content 
fix: non existing innerWidth property with the clientWidth property 
feat: comment mobile breakpoint in order to avoid grid layer removal and items stack with no order

# Description

refactor: some redundant styles and elements for the editor widget content 
fix: non existing innerWidth property with the clientWidth property 
feat: comment mobile breakpoint in order to avoid grid layer removal and items stack with no order

The problem about the reorder of items inside the tab widget is this property:
image

We had it with the default value, but I think that we dont want to trigger the mobile breakpoint in any use case as we want to keep the grid layout always, that's why is set to 0

## Useful links

- [Link to ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/82624)

## Type of change

# How Has This Been Tested?

Please editor widget look and feel still the same
image

No swap of inner widgets on tab widget shrink:
https://github.com/ReliefApplications/ems-frontend/assets/123092672/8a1486b2-fba2-4030-a5fe-11ff3424d36d

## Screenshots

# Checklist:

( * == Mandatory ) 

# More explanation


estelafs (approved):
Everything is working as expected.

libs/shared/src/lib/components/widget-grid/widget-grid.component.ts
```diff
@@ -225,7 +225,7 @@ export class WidgetGridComponent
       pushDirections: { north: true, east: true, south: true, west: true },
       disableScrollHorizontal: true,
       setGridSize: true,
-      mobileBreakpoint: 640,
+      mobileBreakpoint: 0,
```
AntoineRelief: @unai-reliefapp 
everything seems to work fine, but I don't agree with this change

AntoineRelief:
**TODO**

see this comment:
https://github.com/ReliefApplications/ems-frontend/pull/2229/files#r1444490562

it seems that when on mobile ( or when size of the container is < than 640 ), we enter mobile mode ( which is mode, otherwise, the widgets are unreadable, as they are too small ).
nevertheless, the order of the widgets is not correct
in fact it seems that mobile mode uses the order of the widgets in the array, regardeless of the position of the widgets
one solution could be to change the order of the widgets, to match with the one _we see_.

we should first see widgets on top left then top right etc
so ordering should first be done on the 'y', then on the 'x'

guisoares1:
What was done:

- I did what was requested in this [comment](https://github.com/ReliefApplications/ems-frontend/pull/2229/files#r1444490562).
- I sorted based on the rule provided.
-----
Video

https://github.com/ReliefApplications/ems-frontend/assets/55603259/255b2ce6-c8fd-4439-a11f-bfcd76dacfb3

libs/shared/src/lib/components/widget-grid/widget-grid.component.ts
```diff
@@ -118,14 +118,15 @@ export class WidgetGridComponent
   }
 
   ngOnInit(): void {
+    this.widgets.sort((a, b) => a.y - b.y || a.x - b.x);
```
AntoineRelief: @guisoares1 
please do a quick test, where you would load a dashboard with many widgets, add a new widget, put the widget on top left, then resize the screen to see if the widget is still on top or at bottom

all these actions without refreshing the dashboard, so the instance of the widget grid is still the same

I'm asking you to do this test, because I'm not sure this use-case is covered.
indeed, you only sort on init, but the list can change dynamically when editing the dashboard

guisoares1:
@AntoineRelief I did this test below, if I forgot something I can do some more tests:

https://github.com/ReliefApplications/ems-frontend/assets/55603259/e86a0897-f238-43d5-ab44-97254789e629

AntoineRelief:
@guisoares1 
looks good to me 👍

PR #2249, opened 2024-01-04
feat/AB#72599_improve_history_section_of_record

RenzoPrats:
# Description

Improved history section of record: 

- changes on date range should automatically update the view
- changes on field should automatically update the view
- change label "select a field" to be "select fields", as it is a multi select
- Add clear date button in right of "date range"
- hide the cross icon in the field selector when no field selected
- put the cross icon in red

## Useful links

[ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/72599)

## Type of change

# How Has This Been Tested?

It has been tested accessing some history of a record, trying to filter by only date and verifying if the clear button and ui is working fine.

## Screenshots

Peek 04-01-2024 12-47

# Checklist:

( * == Mandatory ) 


libs/shared/src/lib/components/record-history/record-history.component.ts
```diff
@@ -174,6 +174,25 @@ export class RecordHistoryComponent
     } else {
       setSubscription();
     }
+
+    this.filtersDate.valueChanges.subscribe(() => {
```
AntoineRelief: @RenzoPrats 
don't forget about adding takeUntil !

libs/shared/src/lib/components/record-history/record-history.component.ts
```diff
@@ -174,6 +174,25 @@ export class RecordHistoryComponent
...
+      const startDate = this.filtersDate.get('startDate')?.value
+        ? new Date(this.filtersDate.get('startDate')?.value as any)
```
AntoineRelief: @RenzoPrats 
this is not working, as, if you select fields, then change the date range ,the filter on the field is being removed
in addition, it's a duplication of code, because we have exactly the same thing a few lines below

what you should do instead, is to change the logic so this method: applyFilter, does not exist anymore, and instead, the fields are also part of the reactive form
so we can just subscribe to changes on the form, and apply logic based on that, taking all parameters into account

unai-reliefapp:
This styling could be improved right? Is a little bit Picassian
image

Also I guess this is out of the scope, but to be sure, the time set in the excel file from history contains the timezone, that should be enough to avoid any mismatch confusion from the end user between time display in the history sidenav and the file

image

PS: Push some duplicate code removal 😉 

KR, Unai

AntoineRelief:
@RenzoPrats 
Don't do anything about timezone, there is another discussion for that

And put à justify-between for the row, so the button is on the right

Cc @unai-reliefapp

AntoineRelief:
@RenzoPrats 
cc @unai-reliefapp 
all good 👍

PR #2251, opened 2024-01-04
Fix/ab#82564 widgets unnecessarily reloading

NathanHGit:
# Description

Widgets are unnecessarily reloaded when dashboard filters are used. Adding a regex expression to prevent unwanted reloading. 

For the maps, I added the contextFilters variable to the layer to be able to read them from the map component. I didn't find a better way of doing it. 

I was looking at `featureSatisfiesFilter()` into `layer.ts` during my investigation and it may be unused. It is called if `this.filter` isn't empty or when it is initialized: `this.filter = get(options, 'filter', DEFAULT_LAYER_FILTER);` it seems that `options.filter` doesn't exist, then `DEFAULT_LAYER_FILTER` is used. I don't know all the features of the project that's why I didn't remove this part.

## Useful links

- [Please insert link to ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/82564/)

## Type of change

# How Has This Been Tested?

Setup: I created a dashboard and added each widget in it. I created a context filter.
Test: I tried to modify the dashboard filter to see if widgets were reloaded or not. I did the same test with filter configuration for each of them.

## Screenshots

peek
In this example, the pie chart widget is the only one configured with context filters.

# Checklist:

( * == Mandatory ) 

# More explanation

### DIFF
--- a/libs/shared/src/lib/components/aggregation/aggregation-grid/aggregation-grid.component.ts
+++ b/libs/shared/src/lib/components/aggregation/aggregation-grid/aggregation-grid.component.ts
@@ -110,11 +110,13 @@ export class AggregationGridComponent
   }
 
   ngOnInit(): void {
-    this.contextService.filter$
-      .pipe(debounceTime(500), takeUntil(this.destroy$))
-      .subscribe(() => {
-        this.getAggregationData();
-      });
+    if (this.contextService.filterRegex.test(this.contextFilters as string)) {
+      this.contextService.filter$
+        .pipe(debounceTime(500), takeUntil(this.destroy$))
+        .subscribe(() => {
+          this.getAggregationData();
+        });
+    }
     this.queryBuilder.isDoneLoading$.subscribe((doneLoading) => {
       if (doneLoading) {
         this.getAggregationFields();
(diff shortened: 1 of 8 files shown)


unai-reliefapp:
Hi @NathanHGit , just wondering, if a widget contains some contextFilters in their settings, and we then delete them from the settings would it be triggered because it's already subscribed previously or the widget state is reset?

And maybe same on the other way around, if no contextFilters set and then setting them

Also contextFilters can contain fix values, without the need of {{filter.}} notation
image

Would it take them in account?

KR, Unai

NathanHGit:
@unai-reliefapp for you first question, it works because the widget is reloaded after saving the settings. Here is a demonstration: 

peek

Concerning the second one, it works as well:

image

Both show the same data but I added this filter on the left map:

image

According to my knowledge, this kind of filter applies once when loading and then never change.

unai-reliefapp:
> Concerning the second one, it works as well:
> 
> image
> 
> Both show the same data but I added this filter on the left map:
> 
> image
> 
> According to my knowledge, this kind of filter applies once when loading and then never change.

I was concerned because of the regex: https://github.com/ReliefApplications/ems-frontend/blob/a520dfd6f126b16020c6c496668d84133aebeb2b/libs/shared/src/lib/services/context/context.service.ts#L48

As those fix values may not contain the {{filter.}} notation and the regex would be falsy not triggering the subscription even if the widget has contextFilters.

If it works fine 👍

AntoineRelief:
@NathanHGit 
cc @unai-reliefapp 
looks good to me, will now try live with WHO applications 👍

PR #2272, opened 2024-01-10
Fix/ab#83183 default value not working with ref data

NathanHGit:
# Description

When using reference data for a question, the default value popup was not working correctly. The data was being stored in `.visibleChoices` instead of `.choices` to prevent the loaded data from appearing in the choices tab. Unfortunately, the default value pop up is based on `.choices` and as a result, nothing was displayed. It is a surveyjs feature and it cannot be replaced.

While there are no issues if reference data choices appear in the choices tab, it may not be very intuitive. We looked for a solution with Matthis. Here are the results:
- Allow only one "choices" tab expansion at a time
- By default, the current one will be opened and won't be closable
- When clicking on other "choices" tab, previous values will be cleared
- A confirmation pop-up will inform the user of this behavior

SurveyCreator and PanelModel don't have an event that listens for expanding or collapsing tabs, so I used `PanelModel.onPropertyChanged`. Since I couldn't listen before expanding a tab, I had to collapse it again. This approach may seem less than ideal, but I haven't found a better solution.

I moved the `new SurveyModel` of the filter-builder-modal above the `renderGlobalProperties()` because reference data weren't loaded.

## Useful links

- [Please insert link to ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/83183/)
- [PanelModel.onPropertyChanged](https://surveyjs.io/form-library/documentation/api-reference/panel-model#onPropertyChanged)

## Type of change

# How Has This Been Tested?

Bug fix: I created a form with questions linked to reference data and checked if the default value modal was loading correctly.
New feature: I created a form with a dropdown question and switched between choices tabs to control the behavior.

## Screenshots

Bug fix

peek

New feature

peek2

# Checklist:

( * == Mandatory ) 

# More explanation

### DIFF
--- a/libs/shared/src/lib/survey/global-properties/reference-data.ts
+++ b/libs/shared/src/lib/survey/global-properties/reference-data.ts
@@ -242,10 +242,8 @@ export const render = (
             filter
           )
           .then((choices) => {
-            question.choices = [];
-            // this is to avoid that the choices appear on the 'choices' tab
             question.setPropertyValue(
-              'visibleChoices',
+              'choices',
               choices.map((choice) => new ItemValue(choice))
             );
             // manually set the selected option (not done by default)
--- a/libs/shared/src/lib/survey/global-properties/reference-data.ts
+++ b/libs/shared/src/lib/survey/global-properties/reference-data.ts
@@ -242,8 +242,9 @@ export const render = (
             filter
           )
           .then((choices) => {
+            question.choices = [];
             question.setPropertyValue(
-              'choices',
+              'visibleChoices',
               choices.map((choice) => new ItemValue(choice))
             );
             // manually set the selected option (not done by default)
--- a/libs/shared/src/i18n/en.json
+++ b/libs/shared/src/i18n/en.json
@@ -807,6 +807,10 @@
 			}
 		},
 		"formBuilder": {
+			"choicesTabAlert": {
+				"confirmationMessage": "The opening of this new tab will erase your current choices.\nAre you sure you want to continue and clear your configuration?",
+				"expandTab": "Tab change"
+			},
 			"errors": {
 				"duplicatedRelatedName": "Duplicated related name for question {{question}} on page {{page}}. Please provide a unique related name (snake case format) to save the form.",
 				"incorrectJSON": "JSON is invalid in JSON Editor, changes haven't been saved",
--- a/libs/shared/src/i18n/fr.json
+++ b/libs/shared/src/i18n/fr.json
@@ -818,6 +818,10 @@
 			}
 		},
 		"formBuilder": {
+			"choicesTabAlert": {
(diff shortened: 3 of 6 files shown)


AntoineRelief:
@NathanHGit @Matthis-M-ReliefApps 
I think we could have a look at how questions behave when using choices from restful service, because for this kind of configuration I don't think we use the "choices" field, but the default value can be set

libs/shared/src/lib/survey/global-properties/reference-data.ts
```diff
@@ -242,10 +242,8 @@ export const render = (
             filter
           )
           .then((choices) => {
-            question.choices = [];
-            // this is to avoid that the choices appear on the 'choices' tab
             question.setPropertyValue(
-              'visibleChoices',
+              'choices',
```
AntoineRelief: @NathanHGit 

does this line save the choices in the DB?

cc @Matthis-M-ReliefApps

NathanHGit: Yes it does, I will add code to empty the choices before saving the form.

AntoineRelief:
@NathanHGit 
another comment
where did you find the default popup would use only "choices"?

NathanHGit:
@AntoineRelief I didn't really find it, it's more something I deduced. I searched through the doc and various forums, but it doesn't seem possible. I didn't express myself well in the pull request, sorry

AntoineRelief:
there may be something to do with 
```
this.surveyCreator.onPropertyGridShowModal.add(
      (sender: any, options: any) => {
        // and there, put some code
      }
    );
```

I tested it quickly, it seems it can be quite similar to what we have done there, for example in the form-builder.component file:

```
    this.surveyCreator.survey.onAfterRenderQuestion.add(
      renderGlobalProperties(this.referenceDataService) as any
    );
    this.surveyCreator.survey.onAfterRenderQuestion.add(
      this.formHelpersService.addQuestionTooltips as any
    );
```

I still don't understand exactly to which element the similar logic should be applied, but my guess is that the custom properties we build ( tooltip / ref data ) are not available in the modal

NathanHGit:
The last push isn't an improvement but tests.

As pointed out before, using the "choices" property causes the data to be saved in the database. Data is duplicated which is not a viable solution. To prevent it we can clear "choices" before saving the survey. I haven't implemented it.

As proposed by Antoine, I tried to use [onPropertyGridShowModal](https://surveyjs.io/survey-creator/documentation/api-reference/survey-creator#onPropertyGridShowModal). I couldn't make it work but it seems possible. 

The objects given by the function are spreaded out, and I couldn't find the correct property to replace. At best, it only worked after reopening the popup. According to the documentation (which I double checked), the event is called before the pop up loads. However, in my case, the replaced data wasn't taken into account directly. I concluded that I was dealing with the wrong property.

AntoineRelief:
closing this one as it does not solve

PR #2274, opened 2024-01-10
fix/AB#83212_filtering_with_multidropdown_reference_data_not_work

RenzoPrats:
# Description

Fix: Filtering with multi-dropdown reference data does not work.
:warning: I'm not sure if it's a issue but this is only working when we have the field to filter as multi drop down too(like the filter), otherwise the filter in the backend causes an error.

## Useful links

[ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/83212)

## Type of change

# How Has This Been Tested?

It has been tested creating a dashboard filter with a multi-select dropdown reference data, unticking 'Is primitive value' from the 'Choices from reference data', creating  filter on a widget with "{{filter.'name'.'field'}}" notation.

## Screenshots

Peek 10-01-2024 17-25

# Checklist:

( * == Mandatory ) 


estelafs:
> ⚠️ I'm not sure if it's a issue but this is only working when we have the field to filter as multi drop down too(like the filter), otherwise the filter in the backend causes an error.

Only the multi-select dropdown with the untick 'Is primitive value' is working and the normal dropdown with the untick 'Is primitive value' is another ticket in progress, so no problems about this here.

libs/shared/src/lib/services/context/context.service.ts
```diff
@@ -218,13 +219,24 @@ export class ContextService {
...
     const contextRegex = /(?<={{context\.)(.*?)(?=}})/gim;
-
     if ('field' in filter && filter.field) {
       // If it's a filter descriptor, replace value ( if string )
       if (filter.value && typeof filter.value === 'string') {
         const filterName = filter.value?.match(filterRegex)?.[0];
         if (filterName) {
-          filter.value = get(this.availableFilterFieldsValue, filterName);
+          const firstFilterName = filterName.split('.')[0];
+          const secondFilterName = filterName.split('.')[1];
```
estelafs, line 287: @RenzoPrats You  forgot to check if the secondFilterName also exists, otherwise for multi-select dropdown with the 'Is primitive value' and multi-select dropdown with normal static choices arrays the filter does not work

estelafs (approved):

AntoineRelief:
I changed my mind about this one

In fact, when we have a non-primitive tagbox (ref data) , the value provided to filters should be something like that for example

if user select France & Brazil, in a dashboard filter, with a "country" field

then data sent by the form would be something like:
[{ text: "France", value: obj }, { text: "Brazil", value: obj }]

but the data the widgets would be able to use should be something like:

{
  country: {
   code: ["fra", "bra"],
  name: ["france", "Brazil"],
...  
}
}

so we build a json, adding one property per properties we find in the selected values ( there, I only put one level, but it can be more in some cases )

MwanPygmay:
@RenzoPrats is it possible to filter by several values? It seems to me this does not work:
multi-select-bug


EDIT: it is out of the scope of the ticket, and will be addressed in https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/84023

RenzoPrats:
@MwanPygmay I was unsure about that too but the value passed to the filter is for example "episode": ["Anatomy park", "pilot"] so it will try to filter a record that have both values not a record that contains at least one of these values, perhaps we could change the way filters are working in this part.

PR #2291, opened 2024-01-15
fix/AB#83434_select_menu_disabled_input_does_not_work

RenzoPrats:
# Description

Fix: ui-select-menu disabled input doesn't work, even if set to true, the select menu is never disabled.

## Useful links

[ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/83434)

## Type of change

# How Has This Been Tested?

Configured any ui-select-menu input "disabled" as true and verified if it's working correctly.

## Screenshots

Captura de tela de 2024-01-15 17-59-58

# Checklist:

( * == Mandatory ) 

### DIFF
--- a/libs/ui/src/lib/select-menu/select-menu.component.ts
+++ b/libs/ui/src/lib/select-menu/select-menu.component.ts
@@ -284,7 +284,8 @@ export class SelectMenuComponent
    * @param isDisabled is control disabled
    */
   public setDisabledState(isDisabled: boolean): void {
-    this.disabled = isDisabled;
+    // Check if the input is explicitly set, otherwise use the value passed to setDisabledState
+    this.disabled = this.disabled !== undefined ? this.disabled : isDisabled;
   }
 
   /**


libs/ui/src/lib/select-menu/select-menu.component.ts
```diff
@@ -284,7 +284,8 @@ export class SelectMenuComponent
    * @param isDisabled is control disabled
    */
   public setDisabledState(isDisabled: boolean): void {
-    this.disabled = isDisabled;
+    // Check if the input is explicitly set, otherwise use the value passed to setDisabledState
+    this.disabled = this.disabled !== undefined ? this.disabled : isDisabled;
```
AntoineRelief, line 288: @RenzoPrats 
cc @NathanHGit 
if you set disabled, then you can't update it anymore no?
as you will always reuse previous value
it looks weird to me

NathanHGit: @RenzoPrats
I was investigating the same thing. The fix allows the component to be properly initialized but it does not seem dynamic. I tried to find a place where this might be the case but couldn't find one.

RenzoPrats: @AntoineRelief @NathanHGit This issue is weird, for some reason this setDisabledState from ControlValueAccessor was setting false to the disabled all the time, even when we set disabled input as true in the html, it was modifieng to false. I'm checking this case you've mentioned

AntoineRelief: @RenzoPrats 
because the control is not disabled
so this is taking control over the disabled state of the component
please jump to something else, this is really not a priority

PR #2315, opened 2024-01-22
fix/AB#84058-replaceFilter-method-contextService-casts-objects-to-strings

estelafs:
# Description
Context service:
-  Updated filterRegex propriety to handle matches like `"{{filter.field}}" ` and not just `{{filter.field}}`
- Updated replaceFilter() method  to match both new possible filters placeholders and to return the fieldValue stringified

For every call of the contextService.shouldRefresh method in all the widgets:
- referenceDataVariableMapping is saved as a string, so on applying the JSON.stringify, replace and then JSON.parse In the replaceFilter() method was never able to get the correct value. 
- to fix that and also patronize the contextFilters to the new replaceFilter() method, the contextFilters and referenceDataVariableMapping are parsed into objects before the shouldRefresh used it so the replaceFilter() can work as it should. The graphQLVariables() was doing this before call the contextService.replaceFilter, but  the shouldRefresh() method was not

## Useful links

- Please insert link to ticket: https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/84058

## Type of change

# How Has This Been Tested?
Set dashboard with filters or context dashboards to dynamically change the contextFilters or referenceDataVariableMapping of graphql reference data to test the filters with all the possible fields (arrays, booleans, strings etc)

## Screenshots
Using arrays/objects:
array-obj
after-before-array-obj

Booleans:
boolean
after-before-boolean

# Checklist:

( * == Mandatory ) 

# More explanation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,27 @@
+# [2.6.0](https://github.com/ReliefApplications/ems-frontend/compare/v2.5.0...v2.6.0) (2024-01-23)
+
+
+### Bug Fixes
+
+* could not edit filter from dashboard-filter html ([43c37d0](https://github.com/ReliefApplications/ems-frontend/commit/43c37d059514378cd3a22d3e56b8c9ca1a804e60))
+* dashboard buttons would not be correctly saved ([9b32345](https://github.com/ReliefApplications/ems-frontend/commit/9b3234510402a69198381d4356da91b11aca5514))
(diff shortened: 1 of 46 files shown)


RenzoPrats (approved):
Working fine, looks good to me!

AntoineRelief:
@estelafs 
I simplified a bit the code
I didn't quite like all the part about creating an object you would then pass to the contextService.shouldRefresh

instead, I built a method in the service, that will transform the object into one, that will use objects instead of objects saved as strings ( what we have for contextualFilters & refData ) whenever this is doable

PR #2325, opened 2024-01-25
fix/AB#83183_default-value-modal-with-choices-from-ref-data

matheus-relief:
# Description

The issue was the the dropdown with the choices was populated with the choices array of the question. However, that was never set and as far as I can tell (and from the comment there) the only reason for it was to hide the choices in the choices tab. So, instead of doing that, I'm showing the property conditionally, so it doesn't show on the settings if you have a reference data selected.

Another problem, specifically for the dashboard filter builder, was just the order of the code.

## Useful links

- Please insert link to ticket: https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/83183

## Type of change

# How Has This Been Tested?

See below

## Screenshots
Peek 2024-01-23 14-18

# Checklist:

( * == Mandatory ) 

# More explanation


NathanHGit:
@matheus-relief It works fine, but choices are saved in the database. Perhaps we could add an event that clears `.choices` before saving the form. Also, because we desactivate the choices, they don't appear into the corresponding tab. In my opinion, it is not very intuitive, maybe we can add a warning message.

image

AntoineRelief:
@matheus-relief 
😍 incredible

AntoineRelief:
@NathanHGit 
good point indeed

AntoineRelief:
@NathanHGit 

with last push, choices should not be saved, and it should be possible to select a default value for the reference data

now, only issue i see is that ( and it doesn't affect tagbox, just dropdown it seems ), if you have a default selected value, and you open the form builder, it won't appear:

Capture d’écran 2024-01-25 à 14 23 36

Capture d’écran 2024-01-25 à 14 21 39

but if you edit it again, you would see it in the form builder, but not anymore in the popup after opening it again

Capture d’écran 2024-01-25 à 14 21 53


cc @matheus-relief

NathanHGit:
@AntoineRelief I found a solution to the last comment but I'm not entirely satisfied with it. They were two issues. The default value wasn't loaded when opening the form builder; the default value wasn't sent to the popup. 

For the popup, I manually sent the default value like, as it is done for the visibleChoices. 
Now, regarding the loading of the default value, it is a bit more tricky. I don't know why, but the question value can take various formats depending on the origin of the input.

From dashboard filters
```
{
    "name": "France",
    "value": "FR"
}
```

From forms and dashboard filters designer
```
{
    "text": "France",
    "value": {
        "name": "France",
        "value": "FR"
    }
}
```

From forms designer
```
{
    "text": "Germany",
    "value": {
        "pos": {
            "start": 434,
            "end": 489
        },
        "name": "Germany",
        "value": "DE"
    }
}
```

I added conditions to handle each case but it isn't very clean. Additionnaly, depending on the origin of the input, we don't need the same data to display the value. I guess this is due to the survey format.

From dashboard filters
```
{
    "text": "France",
    "value": {
        "name": "France",
        "value": "FR"
    }
}
```

From others
```
{
    "name": "France",
    "value": "FR"
}
```

I also noticed several bugs related to reference data:
- when saving the form from the designer, data is reloaded, but reference data initialization is not called again
- there is a bug with object format when switching between primitive and non-primitive values if a default a value is set
- if you select a value from the dashboard filters, go to the designer, change the default value, save, then the displayed value will be bugged

matheus-relief:
@NathanHGit some changes I made on the last commit:

* Instead of passing the `value` to the modal, we're now passing the `defaultValue`. I think your solution was also technically correct, because we can't change the value of the question from the survey designer, so I guess any value there would need to be the default value.
* I think I managed to simplify a bit the logic to deal with all the different structures of the choices by creating an `ItemValue` from the question value and checking it's id, which from my tests seem to be standardized.
* "Fixed" the issue when we change the boolean for primitive values simply by resetting the default value when we toggle it and updating the `valuePrimitive` of the dropdown that was only set on init before
* Fixed this as well: "if you select a value from the dashboard filters, go to the designer, change the default value, save, then the displayed value will be bugged"

The only thing I wasn't able to fix was the following:
> when saving the form from the designer, data is reloaded, but reference data initialization is not called again

Indeed the `onAfterRenderQuestion` is only called the first time the question is loaded, and I couldn't find another event that is raised on the occasions we want. Maybe we can use the `onAfterRenderPage`, because that seems to be called again and adapt the render function to work with it.

Few tests: 
Peek 2024-01-23 14-18

btw it seems they [fixed the issue](https://github.com/surveyjs/survey-library/issues/7703) with the pos artifacts on a recent issue

cc @AntoineRelief

NathanHGit (approved):
It works fine. All bugs were fixed, and suprisingly, even this one:
> when saving the form from the designer, data is reloaded, but reference data initialization is not called again

I changed the logic to handle dashboard filters (truncated value format)

estelafs:
I tested it thoroughly and everything seems to be working as it should, the bugs were fixed, but I found a new one that I had to fix:

When defining a default value to be used to filter a widget with graphql ref data using graphqlVariables mapping, the default filter value did not work when initializing because the first pageSize was always the default DEFAULT_PAGE_SIZE = 25, and not 1 as it should be to filter one value:
[graphqlVariables.webm](https://github.com/ReliefApplications/ems-frontend/assets/28535394/2cc82396-d695-4989-a00c-921d6b4b36d1)

estelafs (approved):

PR #2326, opened 2024-01-25
AB#84253_filter_not_reloading_when_several_widgets

MwanPygmay:
# Description

Removed global behaviour of regex that causes weird problems. (bug was in fact not related to tabs)

## Useful links

- [Please insert link to ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_boards/board/t/App%20Builder%20-%20Core/Stories/?workitem=84253)

## Type of change

# How Has This Been Tested?

Tried same setup as Matheus, in my case the reverse problem happened (my tabs were before my summary card).

## Screenshots

Why the bug did occur:
Screenshot from 2024-01-25 13-00-47

# Checklist:

( * == Mandatory ) 


libs/shared/src/lib/services/context/context.service.ts
```diff
@@ -58,7 +58,7 @@ export class ContextService {
   /** Is filter opened */
   public filterOpened = new BehaviorSubject<boolean>(false);
   /** Regex used to allow widget refresh */
-  public filterRegex = /["']?{{filter\.(.*?)}}["']?/g;
+  public filterRegex = /["']?{{filter\.(.*?)}}["']?/;
   /** Regex to detect the value of {{filter.}} in object */
   public filterValueRegex = /(?<={{filter\.)(.*?)(?=}})/gim;
```
AntoineRelief, line 66: @MwanPygmay I guess we have the same problem for everything no?
filterValueRegex + contextRegex as well?

AntoineRelief (changes requested):

libs/shared/src/lib/services/context/context.service.ts
MwanPygmay: yeah that makes sense I'll update the PR

MwanPygmay: Nevermind they are only used with 'match' that causes no problem, and as we use replace with the contextRegex it is essential to keep the g flag

AntoineRelief: ok
also note that the filterRegex is used not only to test if the widget should use contextService, but also inside the contextService & map, to listen to geographic extent
please make sure all that will still work

MwanPygmay: checked and it should not cause any problem because: 
- it is either used with .test => that is the problem this ticket is fixing (could also occur in chart, grid...)
- or used with .match but the list returned is never used as a list, just as a bool, so it basically is like a .test

PR #2327, opened 2024-01-25
Revert "AB#84253_filter_not_reloading_when_several_widgets"

AntoineRelief:
Reverts ReliefApplications/ems-frontend#2326


AntoineRelief:
@MwanPygmay 
I had to revert it, it seems indeed to be breaking stuff
some of my widgets don't get any filter update after I merged your code

AntoineRelief:
@MwanPygmay 
okay, found the issue
replaceFilter requires to cast the regexp with g flag it seems

Like that: 
    const replaced = toString.replace(
      new RegExp(this.filterRegex, 'g'),

MwanPygmay:
@AntoineRelief did you fix it? Maybe instead of removing the g flag we could just change the test for a match like done in some other widgets?

PR #2384, opened 2024-02-13
Ab#85564

AntoineRelief:
# Description

Be able to inject graphql variables while using reference data in forms ( or dashboard filter )

https://www.loom.com/share/d91cafd9272c4d498bd90a24fba48434?sid=db8a4ead-d035-4c6b-adb6-d48ba52c46e9

Missing:
- how to clear values when they should disappear ( for example, selection of a region should clear already selected countries out of the region as they are not in the query results we will get for countries )
- non-primitive fields
- test with tagbox & dropdown
Name of the resource in oort-dev.oortcloud.tech is RegionCountryAdmins

Variables to setup for the countries field is:

{ "filter": { "regionid_in": "{{regions}}" } }

admin1:

{ "filter": {"countryid_in": "{{countries}}“ } }

admin2:

{"filter": {"admin1id_in": "{{admin1s}}" } }

## Useful links

- Please insert link to ticket
- Please insert link to back-end branch if any
- Please insert any useful link ( documentation you used for example )

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Screenshots

# Checklist:

( * == Mandatory ) 

# More explanation


unai-reliefapp:
@AntoineRelief Here is the update of removal of selected values and the related values removal(sorry i forgot to mute the mic, maybe you'll see some cofs)

https://github.com/ReliefApplications/ems-frontend/assets/123092672/82ac4cfa-6f88-4320-a1b7-84120d950183

AntoineRelief:
@unai-reliefapp 
nice
and if you remove regions, but you have some selected countries, will they remain? ( they should )

unai-reliefapp:
> { "filter": {"countryid_in": "{{countries}}“ } }

No, they way i have to do it is check if it has a related filter set and if no value for it is set in the survey data then clear it, that's why works in cascade.

But it does not take in account any special case, for that we need another approach. Do you have something in mind for that? A new property in the  tagbox/dropdown inside reference data?

AntoineRelief:
> > { "filter": {"countryid_in": "{{countries}}“ } }
> 
> No, they way i have to do it is check if it has a related filter set and if no value for it is set in the survey data then clear it, that's why works in cascade.
> 
> But it does not take in account any special case, for that we need another approach. Do you have something in mind for that? A new property in the  tagbox/dropdown inside reference data?

@unai-reliefapp 
How I'd see it working
This form is just an example so there may be different cases, but I think this one is the most flexible

- regions is cleared by user, then countries is filed with all countries from api ( no filtering ) and choices are not removed 
- countries is cleared, then admin2s values should be set by surveyjs trigger rules to blank & be hidden ( no code needed I think, it's configuration )
- same for admin 1 & admin 2

libs/shared/src/lib/survey/global-properties/reference-data.ts
unai-reliefapp, line 1: Is there any actual use for these lines in this file?
```
    // Prevent selected choices to be removed when sending the value
    question.clearIncorrectValuesCallback = () => {
      // console.log(question.visibleChoices);
      // console.log(question.value);
    };
```

AntoineRelief: @unai-reliefapp 
I think it's there because without them, some default logic would apply that could cause performance issues

AntoineRelief: but it may be linked to what we're trying to do in fact

PR #2409, opened 2024-02-22
AB#86500_add_valid_json_validator

MwanPygmay:
# Description

Add validator to check for invalid JSON value.
Add alert to warn user if the value is invalid.

## Useful links

- [Please insert link to ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/86500)

## Type of change

# How Has This Been Tested?

tested in summary card, tried with empty json, normal and with errors.

## Screenshots

validJsonValidator

# Checklist:

( * == Mandatory ) 

# More explanation

### DIFF
--- a/libs/shared/src/lib/components/widgets/common/graphql-variables-mapping/graphql-variables-mapping.component.html
+++ b/libs/shared/src/lib/components/widgets/common/graphql-variables-mapping/graphql-variables-mapping.component.html
@@ -31,6 +31,9 @@
     'components.widget.settings.summaryCard.card.dataSource.layoutStyles.queryFilters.help'
       | translate
   }}</ui-alert>
+  <ui-alert variant="danger" *ngIf="!control.valid">{{
+    'common.errors.invalidJson' | translate
+  }}</ui-alert>
   <div
     *ngIf="availableQueryVariables.length > 0"
     mwlResizable
--- a/libs/shared/src/lib/components/widgets/common/graphql-variables-mapping/graphql-variables-mapping.component.html
+++ b/libs/shared/src/lib/components/widgets/common/graphql-variables-mapping/graphql-variables-mapping.component.html
@@ -31,7 +31,7 @@
     'components.widget.settings.summaryCard.card.dataSource.layoutStyles.queryFilters.help'
       | translate
   }}</ui-alert>
-  <ui-alert variant="danger" *ngIf="!control.valid">{{
+  <ui-alert variant="danger" *ngIf="control.errors?.invalidJSON">{{
     'common.errors.invalidJson' | translate
   }}</ui-alert>
   <div
--- a/libs/shared/src/lib/components/widgets/common/graphql-variables-mapping/graphql-variables-mapping.component.ts
+++ b/libs/shared/src/lib/components/widgets/common/graphql-variables-mapping/graphql-variables-mapping.component.ts
@@ -63,7 +63,8 @@ export class GraphqlVariablesMappingComponent implements OnInit, OnChanges {
   public style: any = {};
 
   ngOnInit() {
-    this.control.setValidators(validJson());
+    this.control.addValidators(validJson());
+    this.control.updateValueAndValidity();
   }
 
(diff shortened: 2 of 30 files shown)


unai-reliefapp (commented):
Hi @MwanPygmay , maybe set the control validator inside the graphql variables component in order to keep validator and related control tight coupled?
We want the validjson feature in the graphql variables component always right?
Place it there and we don't have to go control by control to set the validator

What do you think?
KR, Unai

cc @AntoineRelief

libs/shared/src/lib/components/widgets/common/graphql-variables-mapping/graphql-variables-mapping.component.ts
```diff
@@ -55,6 +62,10 @@ export class GraphqlVariablesMappingComponent implements OnChanges {
   /** size style of editor */
   public style: any = {};
 
+  ngOnInit() {
+    this.control.setValidators(validJson());
```
unai-reliefapp:
```suggestion
    this.control.setValidators(validJson());
    this.control.updateValueAndValidity();
```
https://angular.io/api/forms/AbstractControl#setValidators
Also set validators would replace any other validator set in the control, maybe would be better to use the `addValidators()` method?

KR, Unai

unai-reliefapp (changes requested):

MwanPygmay:
@unai-reliefapp fully agree with you, made a commit to remove validators from parents and instead setting it from the graphql-mapping component onInit.

libs/shared/src/lib/components/widgets/common/graphql-variables-mapping/graphql-variables-mapping.component.html
```diff
@@ -31,6 +31,9 @@ <h3 class="mb-0">
     'components.widget.settings.summaryCard.card.dataSource.layoutStyles.queryFilters.help'
       | translate
   }}</ui-alert>
+  <ui-alert variant="danger" *ngIf="!control.valid">{{
```
unai-reliefapp: If the control is invalid for any other reason than the `invalidJson: true` this message would also be triggered 😅 
Maybe the `*ngIf` check should be more strict, if no valid and errors contains `invalidJson` 

Besides that if you already check it again and still works fine, approved 👍

unai-reliefapp (changes requested):

libs/shared/src/lib/components/widgets/common/graphql-variables-mapping/graphql-variables-mapping.component.html
MwanPygmay: yup changed it should be all good

unai-reliefapp (approved):

libs/shared/src/lib/components/widgets/common/graphql-variables-mapping/graphql-variables-mapping.component.ts
```diff
@@ -55,6 +62,11 @@ export class GraphqlVariablesMappingComponent implements OnChanges {
   /** size style of editor */
   public style: any = {};
 
+  ngOnInit() {
+    this.control.addValidators(validJson());
```
AntoineRelief: @MwanPygmay @unai-reliefapp 
there is one issue there

I don't think it would affect the performance, but it is not quite clean

indeed, anytime you would enter the component, you would add the validator to the control, so it can end up with a control having multiple times the same validator attached to it

you should better check if the validator is already attached to the control before adding it

putting back as draft, while waiting for a fix

unai-reliefapp: According to Angular that's not a problem
image
https://angular.io/api/forms/AbstractControl#addValidators

AntoineRelief: in that case, i'll merge it, thanks @unai-reliefapp !

PR #2452, opened 2024-03-15
bugfix/AB#88157_SUI-clicking-in-a-point-of-an-expanded-cluster-should-open-the-country-based-on-the-latitude-longitude-of-the-point

unai-reliefapp:
# Description

- feat: check if in cluster layers click event it's propagated from a child marker or just default layer, in the first case we will use the lat lng from the propagated marker, otherwise default event would be taken. This method it's easier than adding click events to all child markers, as we are handling just one event listener per marker groups, and there would be no conflict between cluster click or child marker click

## Useful links

- [Link to ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/88157)

## Type of change

# How Has This Been Tested?

Please refer to video below

## Screenshots

https://github.com/ReliefApplications/ems-frontend/assets/123092672/ed4523e7-d878-4c85-9271-2e602ef7764a

# Checklist:

( * == Mandatory ) 

# More explanation

### DIFF
--- a/libs/shared/src/lib/components/ui/map/layer.ts
+++ b/libs/shared/src/lib/components/ui/map/layer.ts
@@ -880,7 +880,14 @@ export class Layer implements LayerModel {
         // If it's actually one of the features within the cluster, then take that one
         e.propagatedFrom.feature
       ) {
-        const { latitude, longitude } = e.propagatedFrom.feature.properties;
+        const latitude = get(
+          e.propagatedFrom.feature,
+          'geometry.coordinates[1]'
+        );
+        const longitude = get(
+          e.propagatedFrom.feature,
+          'geometry.coordinates[0]'
+        );
         event = { latlng: { lat: latitude, lng: longitude } };
       }
       for (const rule of (map as any)._rules) {
--- a/libs/shared/src/lib/components/ui/map/layer.ts
+++ b/libs/shared/src/lib/components/ui/map/layer.ts
@@ -874,8 +874,17 @@ export class Layer implements LayerModel {
       layer.off('clusterclick', this.rulesListener);
     }
     this.rulesListener = (e) => {
+      let event: any = e;
+      if (
+        get(this.layerDefinition, 'featureReduction.type') === 'cluster' &&
+        // If it's actually one of the features within the cluster, then take that one
+        e.propagatedFrom.feature
+      ) {
+        const { latitude, longitude } = e.propagatedFrom.feature.properties;
+        event = { latlng: { lat: latitude, lng: longitude } };
+      }
       for (const rule of (map as any)._rules) {
-        this.dashboardAutomationService?.executeAutomationRule(rule, e);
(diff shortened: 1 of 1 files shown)


AntoineRelief:
@unai-reliefapp 
nice, will test that right away!

unai-reliefapp:
> @unai-reliefapp nice, will test that right away!

Taking a look at this, a very good improvement would be to trigger the country fetch if current selected country is different from the previous one, because clicking on each feature is triggering same request 😕 

Still need to update some more things @AntoineRelief , please don't merge it yet, as it's working when cluster markers are expanded like that, but if user zooms in and markers are place individually, clicking on them won't work

I'll update full fix soon

libs/shared/src/lib/components/ui/map/layer.ts
```diff
@@ -874,8 +874,17 @@ export class Layer implements LayerModel {
...
     }
     this.rulesListener = (e) => {
+      let event: any = e;
+      if (
+        get(this.layerDefinition, 'featureReduction.type') === 'cluster' &&
+        // If it's actually one of the features within the cluster, then take that one
+        e.propagatedFrom.feature
+      ) {
+        const { latitude, longitude } = e.propagatedFrom.feature.properties;
+        event = { latlng: { lat: latitude, lng: longitude } };
```
AntoineRelief, line 891: @unai-reliefapp 
I'll fix it, but there is something wrong therre
indeed, your logic is based on the fact that the feature would have latitude & longitude properties, but it's not always the case
indeed, the lat & long are coming from 2 fields that the admin can select
it's lucky that for the HR layer you tested they are called latitude longitude, but there is another layer with different fields, and it won't work for this other layer
better use a consistent field like feature.geometry.coordinates

unai-reliefapp: > @unai-reliefapp I'll fix it, but there is something wrong therre indeed, your logic is based on the fact that the feature would have latitude & longitude properties, but it's not always the case indeed, the lat & long are coming from 2 fields that the admin can select it's lucky that for the HR layer you tested they are called latitude longitude, but there is another layer with different fields, and it won't work for this other layer better use a consistent field like feature.geometry.coordinates

Please check my last comment, i'll push the fix, it's not indeed a good fix

AntoineRelief:
@unai-reliefapp 
I just pushed one

if you have anything else to push please let me know

unai-reliefapp:
> @unai-reliefapp I just pushed one
> 
> if you have anything else to push please let me know

It's fine, yes, idk the markers in the cluster where not triggering the rule event on clicking on them, fine for me 👍 
thanks

unai-reliefapp:
> @unai-reliefapp I just pushed one
> 
> if you have anything else to push please let me know

yes, if you zoom in the map and a cluster separates into it's markers, if you click on one, there is no event triggered because the event it's only placed in the cluster

PR #2455, opened 2024-03-21
Feat/ab#86603 ABC SA Attach to record action proceed after record saved

Maximo-reliefapps:
# Description

The quick action button "Attach to record" was being triggered before the popup record edition window showed, making it confusing at UX level, as the user would suppose that clicking on the safe record would trigger the quick button action, and closing the modal would disregard it.

`form-modal-component` has been updated so it has a new property where we can pass an object to overwrite properties of the record. This way we can pass the modified properties from the "Attach to record" action, and save them directly from there. (And see them reflected on the modal).

Following actions are only executed if the form modal is saved.

## Useful links

- [#86603 ABC - SA - Really attach Information to Signal only when update form is saved and closed.](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/86603)

## Type of change

# How Has This Been Tested?

- [x] Check that the records are only attached after confirming the form modal.
- [x] Check that following actions are only executed after confirming the form modal.
- [x] Confirm that the form modal still works everywhere else.

## Screenshots

Recording 2024-03-21 164412

# Checklist:

( * == Mandatory ) 

# More explanation

### DIFF
--- a/libs/shared/src/lib/components/form-modal/form-modal.component.ts
+++ b/libs/shared/src/lib/components/form-modal/form-modal.component.ts
@@ -51,6 +51,7 @@ interface DialogData {
   prefillRecords?: Record[];
   prefillData?: any;
   askForConfirm?: boolean;
+  recordData?: any;
 }
 /**
  * Defines the default Dialog data
@@ -168,6 +169,9 @@ export class FormModalComponent
           })
         ).then(({ data }) => {
           this.record = data.record;
+          if (this.data.recordData) {
+            this.record.data = { ...this.record.data, ...this.data.recordData };
+          }
           this.modifiedAt = this.isMultiEdition
             ? null
             : this.record?.modifiedAt || null;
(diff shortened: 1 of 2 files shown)


Matthis-M-ReliefApps (approved):
Seems to be working very fine, did the following testing:
1. Attach one record, go to second step (record details), cancel. Record is not modified.
2. Attach two records, go to second step, cancel. Record is not modified.
3. Attach one record, go to second step, save and close. Record is attached correctly.
4. Attach two records, go to second step, save and close. Records are attached correctly.

Furthermore, even though we are modifying a core component, since it's only an additive change that requires us to specify "'recordData" at the modal creation, I don't think there should be a risk of regression elsewhere

AntoineRelief:
@Maximo-reliefapps 
all good 👍 
I just refactored it a bit to avoid the dialog1 / dialog2
the feature is working fine

PR #2464, opened 2024-03-28
Feat/ab#88998 keep heatmap scale while zooming

NathanHGit:
# Description

When zooming, the radius of the heatmap stayed fixed to the original value. The data became unreadable when zooming in significantly. I added a coefficient factor that maintains the heatmap radius to the same scale. Each time you zoom in, the map divides into four smaller parts, and then the coefficient increases exponentially (cf leaflet zoom doc). To keep the ratio, we multiply the radius by 2 raised to the power of the zoom value.

## Useful links

- [Please insert link to ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/88998)
- [Leaflet zoom levels](https://leafletjs.com/examples/zoom-levels/)

[Suggested improvement ( not finished )](https://github.com/ReliefApplications/ems-frontend/pull/2477)

## Type of change

- [X Improvement (refactor or addition to existing functionality)

# How Has This Been Tested?

I created a map with several layers that use heatmaps. I tried to zoom in and out to verify the behavior.

## Screenshots

peek (1)

# Checklist:

( * == Mandatory ) 

# More explanation

### DIFF
--- a/libs/shared/src/lib/components/ui/map/utils/leaflet-heatmap.js
+++ b/libs/shared/src/lib/components/ui/map/utils/leaflet-heatmap.js
@@ -141,8 +141,7 @@ L.HeatLayer = (L.Layer ? L.Layer : L.Class).extend({
     }
 
     // Calculate the zoom coefficient to ensure the radius maintains its scale
-    // cf (https://leafletjs.com/examples/zoom-levels/) power of 2
-    const scale = Math.pow(2, this._map.getZoom() - 2) || 1; // -2 because the minimum zoom equals 2
+    const scale = this._map.getZoomScale(this._map.getZoom()) / this._map.getZoomScale(2);
     this._heat.radius(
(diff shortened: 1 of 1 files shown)


Maximo-reliefapps:
@AntoineRelief 
The solution this ticket implements is what I was experimenting with, to modify the heatmap radius property to keep the size on zoom and keep the intensity. Though it's not perfect I don't know if this can be improved on, maybe if we can understand how the colors/intensity is calculated.
From what I understand the colors/intensity are displayed based on the amount of data available, in a no-zoom view the intensity reflects all the data while zooming to an area with fewer data, the intensity shows as the maximum without taking into account the rest of the data.

The other option I tried is to scale the generated heatmap canvas on zooming (instead of regenerating), which works but if it's not generated on a view with all data won't show data outside the area + you will need to calculate the offset of moving the map. For this, you can change the function here `map.on('moveend', this._reset, this)` to use a custom one.

PR #2460, opened 2024-03-25
AB#16252 - Email notifications next

AntoineRelief:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

## Useful links

- [Please insert link to ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/16252)
- [Please insert link to back-end branch if any](https://github.com/ReliefApplications/ems-backend/pull/1005)
- Please insert any useful link ( documentation you used for example )

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Improvement (refactor or addition to existing functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce.

- [ ] Test A
- [ ] Test B

## Screenshots

Please include screenshots of this change. If this issue is only back-end related, and does not involve any visual change of the platform, you can skip this part.

# Checklist:

( * == Mandatory ) 

- [ ] * I have set myself as assignee of the pull request
- [ ] * My code follows the style guidelines of this project
- [ ] * Linting does not generate new warnings
- [ ] * I have performed a self-review of my own code
- [ ] * I have put the ticket for review, adding the oort-frontend team to the list of reviewers
- [ ] * I have commented my code, particularly in hard-to-understand areas
- [ ] * I have put JSDoc comment in all required places
- [ ] * My changes generate no new warnings
- [ ] * I have included screenshots describing my changes if relevant
- [ ] * I have selected labels in the Pull Request, according to the changes with code brings
- [ ] I have made corresponding changes to the documentation ( if required )
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

# More explanation
https://www.loom.com/share/05a716d61b9744faaf51fb304c21d1e5?sid=f87cf896-582a-4f76-93ae-8ceed801b145

PR #2485, opened 2024-04-25
Preview distribution list + More Helpful To, BCC, CC toggles

HasnatL-Adapptlimited:
# Links
[Base PR](https://github.com/ReliefApplications/ems-frontend/pull/2460)

In reference to point 1:
https://github.com/ReliefApplications/ems-frontend/pull/2460#issuecomment-2072440406

# Tickets
- [ ] User can see distribution list on preview page
- [ ] Easier to see, to, cc and bcc at once on the distribution list page


HasnatL-Adapptlimited:
Hey @AntoineRelief,

I was trying to find a way to make ui-chips , labels and paragraph tags more consistent and easy to modify with the rest of the app builder.

With the Ui-Chips it was showing as plain text, is there a class I'm missing, I've used uichiplist and tried to use uiformfielddirective but none of them seemed to make the chips a pill.

libs/shared/src/lib/components/email/steps/select-distribution/select-distribution.component.html
```diff
@@ -11,27 +11,27 @@
...
+      >
+      {{'components.email.distributionList.import.download' | translate}}
+      </ui-button>
+      <!-- Email List Template Button -->
+      <ui-button
+        class="float-right import"
+        icon="cloud_download"
+        category="secondary"
+        variant="primary"
+        uiTooltip="'components.email.distributionList.import.upload' | translate"
```
AntoineRelief, line 29: @HasnatL-Adapptlimited 
this is not valid
it should be: [uiTooltip]
I'll correct that

AntoineRelief:
@HasnatL-Adapptlimited 

merged
Just one comment about the chip
I reverted your changes, but updated the primary chip to have bolder text ( I agree visibility was not good )

primary & secondary were too alike with your proposed changes, while the purpose of these variants are to propose different looks

libs/shared/src/lib/components/email/steps/select-distribution/select-distribution.component.html
HasnatL-Adapptlimited: Yes thats fine will add this to the list for the next few pushes

AntoineRelief: @HasnatL-Adapptlimited 
Please see my comment, I've already corrected it, you don't need to have a look at it

HasnatL-Adapptlimited: Apologies I meant as this has been done in a few places:
image

AntoineRelief: @HasnatL-Adapptlimited 
okay!

Yes in that case it can be fixed
the static ones can stay as they are
but not the translated ones

PR #2492, opened 2024-04-27
AB Body Validator + Future Features Tooltips

HasnatL-Adapptlimited:
# Links
[Base PR](https://github.com/ReliefApplications/ems-frontend/pull/2460)

In reference to point 2 and 4:
https://github.com/ReliefApplications/ems-frontend/pull/2460#issuecomment-2072440406

# Tickets

- [ ] When body is filled, then emptied, the validator doesn't always trigger and we can proceed with empty body.
- [ ] Tooltips for Future features shown

# Next pushes

- [ ] When creating new notification and going to preview step validator snackbar doesnt trigger

### DIFF
--- a/libs/shared/src/lib/components/email/components/dataset-filter/dataset-filter.component.html
+++ b/libs/shared/src/lib/components/email/components/dataset-filter/dataset-filter.component.html
@@ -1,4 +1,4 @@
-<form [formGroup]="query" class="dataset-filter">
+<form [formGroup]="query">
   <!-- Dataset Select -->
   <div class="mt-3 flex gap-5">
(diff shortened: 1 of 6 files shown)


AntoineRelief:
@HasnatL-Adapptlimited 

A bit confuse about this PR
I can see that the status of the PR is ready for review, but I still see there is one push to be done

I suggest to merge this PR as it is, as I can see two things are fixed by it, and you open a new one when your other commit is ready

For future requests, if you don't want us to review them directly, please don't assign us as reviewers, and mark the PR as draft
When all commits are pushed, then you can assign us

cc @Matthis-M-ReliefApps

PR #2599, opened 2024-06-05
Bug/ab#95183 - Sidenav collapsing on form builder firefox

brunotannus:
# Description

Sometimes when resizing some elements in the DOM, specially inside a form builder, the sidenav would shrink until its no longer visible, even when it is set to visible. This issue occurs only on Firefox and can be reproduced by entering a form builder with a resource type question whose grid layout has a lot of fields, which causes the survey builder's width to increase and shrinks the sidenav.
Taking the overflow-y-auto out of the sidenav overflow class vector fixes this issue, but causes the overflow-y of one of its children (`<section>`) to bug out when the sidenav is collapsed, so setting the section's overflow-y to auto fixes this issue.

## Type of change

# How Has This Been Tested?

- Tested on all viewports, zoomed in an out, inside and outside of a form builder, on Chrome and Firefox.

## Screenshots

Before:
2big4u (1)
After:
[screen-recorder-wed-jun-05-2024-15-03-28.webm](https://github.com/ReliefApplications/ems-frontend/assets/39497117/7501198b-a892-4e93-8c7c-ce9ec8fd581f)

# Checklist:

- [x] I have made corresponding changes to the documentation ( if required )
- [x] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/libs/shared/src/lib/components/navbar/navbar.component.html
+++ b/libs/shared/src/lib/components/navbar/navbar.component.html
@@ -4,7 +4,7 @@
     'border-b': !vertical,
     'pt-2 flex-col gap-y-7': vertical
   }"
-  class="flex overflow-y-auto"
+  class="flex"
 >
   <ng-container *ngFor="let group of navGroups; let i = index" class="h-full">
     <!-- Vertical navigation -->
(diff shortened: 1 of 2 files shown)


AntoineRelief:
@brunotannus 

unfortunately I don't believe it's a solution, because it will generate other visual bugs ( as you can see in the video, the main content becomes scrollable, which is something we don't want )

the root cause seems to be the overflow-clip class, which doesn't work as expected on firefox

using overflow-hidden instead solves the issue

I'll push a new commit soon to fix that

brunotannus:
you are right. thanks for clarifying

PR #2617, opened 2024-06-21
AB - Email Notifications - Small Fix

HasnatL-Adapptlimited:
# Description
- DL filtering was changing the dataset filtering before so it has been fixed

## Ticket

- [ ] - DL filtering is no longer affecting the dataset filters

### DIFF
--- a/libs/shared/src/lib/components/email/components/email-template/email-template.component.ts
+++ b/libs/shared/src/lib/components/email/components/email-template/email-template.component.ts
@@ -200,12 +200,12 @@ export class EmailTemplateComponent
       datasetSelect: '',
     });
 
-    this.datasetsForm?.value?.datasets?.forEach((eleDataset: any) => {
-      eleDataset.cacheData = {};
-    });
+    // this.datasetsForm?.value?.datasets?.forEach((eleDataset: any) => {
+    //   eleDataset.cacheData = {};
+    // });
 
     this.selectedEmails = this.emailBackLoad;
-    this.datasets = this.datasetsForm.value.datasets;
+    this.datasets = clone(this.datasetsForm.value.datasets);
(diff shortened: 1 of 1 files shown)


HasnatL-Adapptlimited:
Hey @AntoineRelief 

PR for small fix has been raised!

Thanks
CC: @morgang-adappt @Vaishali1709

libs/shared/src/lib/components/email/components/email-template/email-template.component.ts
```diff
@@ -1028,7 +1028,7 @@ export class EmailTemplateComponent
   getDataSetPreview(filterType: string) {
     this.noEmail.emit(false);
     this.loading = true;
-    const currentDataset = clone(this.selectedDataset);
+    const currentDataset = JSON.parse(JSON.stringify(this.selectedDataset));
```
AntoineRelief, line 1031: @HasnatL-Adapptlimited 

won't clone work there as well?

AntoineRelief: @HasnatL-Adapptlimited

any reason not to use it?
I'm about reviewing the ticket

HasnatL-Adapptlimited: We tried using with clone but realised it was overwriting the data, so we needed a deep clone as to not overwrite it

AntoineRelief: okay, thanks!

AntoineRelief:
@HasnatL-Adapptlimited 

Okay, I'll merge it but I don't see any issue at all with previous code

it would have been good to have a video or repro steps

PR #2644, opened 2024-10-10
feat/AB#104315_EIS-application-path-implementation

unai-reliefapp:
# Description

feat: integrate application shortcut as valid path for frontend 
feat: include resolver for application path in order to fetch correct data for activated route 
feat: add method to extract correct path for give application for the aforementioned reason 
feat: include shortcut value for user query in order to get correct path from given application 
feat: remove useless properties

## Useful links

- [Link to ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/104315)
- [Link to back-end branch](https://github.com/ReliefApplications/ems-backend/pull/1123)
- Please insert any useful link ( documentation you used for example )

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Screenshots

Back office configuration
https://github.com/user-attachments/assets/1098b359-dc74-471f-9108-eec06680c869

Front office behavior
https://github.com/user-attachments/assets/8d2670ac-da9b-426f-941b-a4f553dbf56a

# Checklist:

( * == Mandatory ) 

# More explanation


AntoineRelief:
@unai-reliefapp 

please continue to work on your current ticket, I'm putting info there FYI and to keep notes when I'll go back to it

A few things about this PR:
- it seems the navigation using id is not working anymore in front-end, when a shortcut is configured
- navigation using shortcut is not working in back-office ( not sure it was part of the scope )
- if the request to get the application fails, user only sees a loading screen ( both back-office & front-office )
- when access is not provided / application cannot be found, we can use a similar approach that what we did for the web-widgets integration ( see 'app-error' component in web-widgets application ). We should only redirect the user to a default application if there is no id / shortcut in the url. If there is one but the loading of the app fails, we display the error message
- also, please check that elements ( editor / summary card / grid widgets ) that can build links to other pages, would work with the new url ( I think it's using the id, even if a shortcut is used )

- lastly, I'm putting this one in a different list as it's certainly the most important, you may have done some extra work by doing a specific query to get the shortcut in the router. Instead, you could edit the back-end application.query, to have a new shortcut argument, and find the application based on the shortcut & id ( so the query, in the front-end, would pass id & shortcut, both with the same value, coming from the url, and then, the back-end will find the according application ). It could save a lot of code

unai-reliefapp:
> @unai-reliefapp
> 
> please continue to work on your current ticket, I'm putting info there FYI and to keep notes when I'll go back to it
> 
> A few things about this PR:
> 
> * it seems the navigation using id is not working anymore in front-end, when a shortcut is configured
> * navigation using shortcut is not working in back-office ( not sure it was part of the scope )
> * if the request to get the application fails, user only sees a loading screen ( both back-office & front-office )
> * when access is not provided / application cannot be found, we can use a similar approach that what we did for the web-widgets integration ( see 'app-error' component in web-widgets application ). We should only redirect the user to a default application if there is no id / shortcut in the url. If there is one but the loading of the app fails, we display the error message
> * also, please check that elements ( editor / summary card / grid widgets ) that can build links to other pages, would work with the new url ( I think it's using the id, even if a shortcut is used )
> * lastly, I'm putting this one in a different list as it's certainly the most important, you may have done some extra work by doing a specific query to get the shortcut in the router. Instead, you could edit the back-end application.query, to have a new shortcut argument, and find the application based on the shortcut & id ( so the query, in the front-end, would pass id & shortcut, both with the same value, coming from the url, and then, the back-end will find the according application ). It could save a lot of code

About the last point:

You mean that every time we fetch an application, check that the value is the shortcut and use the related id, or use the default value in the id parameter for the query if it does not find it?

apps/front-office/src/app/app-routing.module.ts
```diff
@@ -1,7 +1,55 @@
...
+/**
+ * Resolve application id
+ * Application can use shortcut values to generate the url
+ * We check the given params for the application route and check if it's a shortcut
+ * If it's a shortcut, return the related application id, if not return current param
+ *
+ * @param route ActivatedRouteSnapshot
+ * @returns resolved application id
+ */
+const applicationIdResolver: ResolveFn<any> = async (
```
AntoineRelief: @unai-reliefapp 

about my last comment, I believe this part is too complex, and you could easily solve it by just sending the id as you get it from the route ( so it's either an id or a shortcut )

then, the back-end must be edited like that: 

const filters = Application.find(accessibleBy(ability).Application)
        .where({ _id: args.id })
        .getFilter();

should become

const filters = Application.find(accessibleBy(ability).Application)
        .where({ or: { id: args.id, shortcut: args.shortcut })
        .getFilter();

there is a lot of errors in what I'm putting there, but that's the overall idea
I may be wrong with this approach

unai-reliefapp: @AntoineRelief  The idea is to replace the shortcut for the id in the route right?
How im going to send the id and shortcut to the backend if i have access to only one of them from the route on every case?
Following your approach im thinking that we can move the logic in the resolver to the application query and always do the check there, e.g.

```
const isIdShortcut = await Application.findOne({
          shortcut: args.id,
        }).select('id');
const filters = Application.find(accessibleBy(ability).Application)
        .where({ _id: isIdShortcut ?? args.id })
        .getFilter();
```
This way we can also avoid to have a resolver for each application, front, back and web widget(now that the scope has been extended)

Im just trying to follow up

AntoineRelief: @unai-reliefapp 
I can't guarantee that would work, but I think the back-end just need to get anything that would identify the application as id argument, would it be the shortcut, or the id. We don't care about that.

Then, you only need to do a query, with a "or" logic, to query both on the id & the shortcut

AntoineRelief: so it's kind of what you're telling, but the isIdShortcut seems an extra step

AntoineRelief: in case you have some issues doing the query on both fields at the same time, you can indeed use your code, or consider using the mongoose id method to check if the id is valid or not

unai-reliefapp: It's working for front and back office, I'll see if i can change the query with that or logic

unai-reliefapp:
Hi @AntoineRelief , do they want the shortcut to be reflected in side urls, e.g.: https://github.com/ReliefApplications/ems-frontend/blob/beta/libs/shared/src/lib/services/download/download.service.ts#L194

Also im assuming to extend the shortcut use to all related places, e.g.
image
If clicking an application and has a shortcut, use it by default, is that correct?

AntoineRelief:
@un

> Hi @AntoineRelief , do they want the shortcut to be reflected in side urls, e.g.: https://github.com/ReliefApplications/ems-frontend/blob/beta/libs/shared/src/lib/services/download/download.service.ts#L194
> 
> Also im assuming to extend the shortcut use to all related places, e.g. image If clicking an application and has a shortcut, use it by default, is that correct?

@unai-reliefapp 
no for the first point, as I believe this is a back-end request so you should not change that

but yes for the front-end navigation, you can use by default the shortcut if we have it

unai-reliefapp:
> @un
> 
> > Hi @AntoineRelief , do they want the shortcut to be reflected in side urls, e.g.: https://github.com/ReliefApplications/ems-frontend/blob/beta/libs/shared/src/lib/services/download/download.service.ts#L194
> > Also im assuming to extend the shortcut use to all related places, e.g. image If clicking an application and has a shortcut, use it by default, is that correct?
> 
> @unai-reliefapp no for the first point, as I believe this is a back-end request so you should not change that
> 
> but yes for the front-end navigation, you can use by default the shortcut if we have it

@AntoineRelief 
That means that by default use the shortcut at any case?
E.g. if i manually use the id to navigate to some application that contains a shortcut, the build up links for widgets would use the shortcut, because it has preference
image
This is acceptable? (I have to navigate using the id manually if shortcut exists for this to happen)

AntoineRelief:
@unai-reliefapp 

yes, it's fine

unai-reliefapp:
Front office error page for application load error
https://github.com/user-attachments/assets/10d98277-61d3-49e7-b0c4-7115347879e8
Back office error page for application load error
https://github.com/user-attachments/assets/7a3b799a-2997-406a-8723-04f84f9ecdd6

@AntoineRelief

AntoineRelief:
@unai-reliefapp 

looks good to me
perhaps we will need to have a button to go back to a default page, but that's out of scope and I need to determine how that would work with WHO, so okay for me like that 👍

unai-reliefapp:
Share feature

https://github.com/user-attachments/assets/8b1938e2-5a32-4269-b75f-7ba135b7b158

All pushed and ready for review @AntoineRelief

apps/front-office/src/app/application/application.component.ts
```diff
@@ -105,15 +104,23 @@ export class ApplicationComponent
     this.applicationService.application$
       .pipe(takeUntil(this.destroy$))
       .subscribe((application: Application | null) => {
+        this.loading = false;
```
AntoineRelief: @unai-reliefapp 
I have a really quick questino
why did you move this?

unai-reliefapp: We dont want the loading spinner neither on successful application load, neither on error right?
Whatever the result, stop loading spinner

PR #2646, opened 2024-10-10
AB: Adappt Email Notification Update

HasnatL-Adapptlimited:
# Description

- [ ] Subscriptions to notifications without admin approval
- [ ] Send separate functionality
- [ ] Saving Distribution List queries as well as static lists
- [ ] Sending resources data (parent-child) in the email
- [ ] Add filter grouping
- [ ] Sending reference data tables in the email

## Useful links

- Backend PR:  [AB: Adappt Email Notification Update](https://github.com/ReliefApplications/ems-backend/pull/1124#top)

## Type of change

- New feature 
- Improvement

## Issues Pending
- Send Individual blocks' email fields must also be in dataset fields (in development)


HasnatL-Adapptlimited:
Hey @AntoineRelief 

PR now open for both frontend and backend

CC: @morgang-adappt, @Nazir-Ahmed-adappt

apps/front-office/src/environments/environment.azure.dev.ts
```diff
@@ -22,20 +21,20 @@ const authConfig: AuthConfig = {
 export const environment: Environment = {
   ...sharedEnvironment,
   production: true,
-  apiUrl: 'https://ems-safe-dev.who.int/api',
-  subscriptionApiUrl: 'wss://ems-safe-dev.who.int/api',
-  frontOfficeUri: 'https://ems-safe-dev.who.int/',
-  backOfficeUri: 'https://ems-safe-dev.who.int/backoffice/',
+  apiUrl: 'https://emspocdev.adapptlabs.com/api',
```
AntoineRelief: @morgang-adappt 

can you revert all changes affecting the environments?

HasnatL-Adapptlimited: @AntoineRelief reverted Environment file changes

HasnatL-Adapptlimited:
Hey @AntoineRelief 

Would there be any possibility to please have this deployed in Dev this week?

CC: @morgang-adappt @Nazir-Ahmed-adappt

AntoineRelief:
@HasnatL-Adapptlimited 
we'll try to deploy it before the end of the week yes

CC: @morgang-adappt @Nazir-Ahmed-adappt

AntoineRelief:
@morgang-adappt 

closing the PR as I merged the changes in `beta-unstable` branch

Note that I'll certainly redo some review work on this feature next monnth, so this is, so far, what I've noticed in back-end & front-end

Now, a few comments about the work in general:
- A lot of JSDoc comments were put as empty, I certainly missed some of them during this first review. As already explained a few months ago, it is mandatory to provide them. Using a linter auto fix will indeed prevent the linter to send errors about them, but it's not a good solution, as the empty are generated empty by the autofix
- Some properties were not correctly typed. I found a lot of typo error or incorrect english naming in the code
- Many artifacts from previous iterations that were not used anymore, such as the import of ng-select, which was removed from the code, but not from the node modules & package.json
- More commenting is always good, especially about elements that require adding new configuration variables, changing existing methods etc.
- Some components didn't follow Angular guidelines. I guess they were not generated using the cli
- many duplications of code. I've solved all of them
- many components that would already exist in the code wouldn't be used, like ui-button, uiFormFieldDirective etc. We've already noticed that in previous iterations

About some specific features, I'm noting that there to be sure it's written somewhere:
- I don't get why there is a blobStorage for the customTemplate, and not for the classic emails. What's the difference between both that would need this specific code to be applied only on the customTemplate?
- I'm pretty sure the existing custom templates are broken, but we will test that after deploying on DEV / UAT as I didn't see any migration that would move the templates from the application model to the new model
- I don't get why we added some code to prevent the edition of the distribution list when editing an email notification

morgang-adappt:
Hey @AntoineRelief, 

I've resolved some of the issues mentioned on the backend (fill empty JSDoc comments, rename `editAndGet` mutations to `edit`, and use img src from blob storage rather than CID attachment universally). PR will be opened once the changes are complete and tested full stack. 

> "I don't get why there is a blobStorage for the customTemplate, and not for the classic emails. What's the difference between both that would need this specific code to be applied only on the customTemplate?"
- Blob storage should be used universally, any references to CID attachments are legacy artifacts that have now been removed. The `EmailNotifications` collection used to contain the email layout as a local field, but to allow reuse for quick-send emails and reduce repetition the `CustomTemplates`  collection now takes this responsibility, and is referenced from the `EmailNotifications` collection. 

> "I'm pretty sure the existing custom templates are broken, but we will test that after deploying on DEV / UAT as I didn't see any migration that would move the templates from the application model to the new model".
- Legacy documents have not yet been migrated, and we do not have a migration script at the moment. Some clarification would be useful on this point; what templates are you referring to from the application model? Templates have moved from the `EmailNotifications` model to `CustomTemplates`, I'd be happy to write a migration script to handle this, as well as to move images stored as base64 strings from MongoDB to blob storage if required. 

> "I don't get why we added some code to prevent the edition of the distribution list when editing an email notification"
 - Since the Distribution List (DL) is commonly used across various email notifications, editing it within a specific email notification is not permitted as it would impact all other notifications where it is used. If users wish to make changes to the DL, they should do so from the DL listing.
 
The other issues such as lack of comments and typos will be worked on. Do you have any other issues we can address in the immediate term?

AntoineRelief:
@morgang-adappt 

thanks for the update! 
About the templates I'm referring to what I believe is now stored in CustomTemplate.
The application I'm concerned about is the `Signal HQ PHI` one.
If you open the application ( back-office, and perhaps front-office ), in UAT environment, you will find this graphQL query `GetApplicationById` and these fields: `distributionLists` & `templates`

At the moment, both arrays are directly stored in the application model in the DB, and should now be stored in the related new models.
I believe the migration for distribution list is quite simple, as the previous model was:
- id -> need to be reused
- name -> need to be reused
- emails -> to transform into "to" field?

for templates, we currently have this model:
- id-> need to be reused
- name-> need to be reused
- type = "email" -> to remove I guess
- content.body -> need to fit to new model
- content.subject -> need to fit to new model

an additional issue about the body, is that if I check the new template model, you don't use anymore the {{dataset}}, but you use {{Block 1}} as you would use it in email notifications, even if the block select will always be empty:

Capture d’écran 2024-10-22 à 15 13 11

for that, I don't what's the easiest.
Either allow to use {{dataset}} and remove this block select, or transform the body in the migration, so {{dataset}} is replaced by {{Block 1}}, and allow the block select to display {{Block 1}} in custom templates

morgang-adappt:
Hey @AntoineRelief 

Thank you very much for the insight! We've started looking into this migration now, are considering the `{{dataset}}`/`{{Block 1}}` conflict, and will begin work on it soon.

PR #2680, opened 2024-11-08
Adappt Email Notification Update Fixes Alpha

AntoineRelief:
# Description

Same PR than this one: https://github.com/ReliefApplications/ems-frontend/pull/2662
But targeting alpha


AntoineRelief:
@morgang-adappt 

I just recreated the PR, targeting alpha, in order to follow the `next` > `beta` > `alpha` pattern

so I'll drop `beta-unstable`, and prefer to use `alpha` now
in case you're doing some work from `beta-unstable`, all the code that was in `beta-unstable` is now in `alpha`, it's just a name change 👍

morgang-adappt:
Hi @AntoineRelief ,

Thank you for the update!

AntoineRelief:
@morgang-adappt 

I think the logic is working fine, as well for the back-end, except for the migrations

if that sounds fine to you, what I can do is merge the 2 branches, but we should consider that the migrations are not okay, so I'd remove them if I do so and we should apply them in another PR

or, we wait for the migrations to be working, as you prefer

AntoineRelief:
@morgang-adappt 

In addition, during the review of the front-end, I noticed that the pagination in the DL tab is not working

indeed, there seems to be a confusion between:
- uniqueDLNames
- cacheDistributionListNames
- cacheDistributionList
- distributionLists

I think distributionLists & cacheDistributionList should be enough, I don't understand why we also have uniqueDLNames & cacheDistributionListNames, because for templates we are just using customTemplates & cachedCustomTemplates and that's working fine

morgang-adappt:
Hi @AntoineRelief ,

We're happy to go ahead with merging the branches without the migrations if you think they're in a suitable state. 

We will have PRs opened for the pagination, buttons for DL management, and the migration script fixes soon. 

Thank you!

PR #2736, opened 2024-12-12
feat/AB#106560_EIS-improve-activity-log-navigation

unai-reliefapp:
# Description

feat: make sure that the application id is fetch if exists on application page enter 
feat: remove not needed method from logger service

## Useful links

- [Link to ticket](https://dev.azure.com/WHOHQ/EMSSAFE/_workitems/edit/106560)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Screenshots

# Checklist:

( * == Mandatory ) 

# More explanation


unai-reliefapp:
@AntoineRelief first view on the application thingie, let me know if the approach could be fine

libs/shared/src/lib/services/application/application.service.ts
```diff
@@ -291,6 +301,7 @@ export class ApplicationService {
         error: () => {
           this.hasErrors = true;
           this.application.next(null);
+          this.loadingApplication.next(false);
```
AntoineRelief, line 304: @unai-reliefapp 
perhaps a finally should do it?

unai-reliefapp: complete function is not triggered if an error is launched, Ive been trying but no luck :(

AntoineRelief: okay then it's good like that

AntoineRelief:
@unai-reliefapp 

it looks quite better like that 👍 

there may be a edge case, and I don't know how often it can happen, but we need to be sure that:
- if you exit an application ( back-office ), then you don't set applicationId in metadata ( so the app service doesn't contain app anymore )
- especially if, from the back-office, you enter an app and quickly go back ( without letting time for the system to load the application )
- if you quickly switch between applications, make sure that the latest application call is taken into account

in fact, you may need to update the apollo call inside the loadApplication method to use a takeUntil, based on the loadingApplication?
so if you emit a new value while loading an app, you know you need to cancel the first apollo query, and only focus on the second one

hope that makes sense!

unai-reliefapp:
> @unai-reliefapp
> 
> it looks quite better like that 👍
> 
> there may be a edge case, and I don't know how often it can happen, but we need to be sure that:
> 
> * if you exit an application ( back-office ), then you don't set applicationId in metadata ( so the app service doesn't contain app anymore )
> * especially if, from the back-office, you enter an app and quickly go back ( without letting time for the system to load the application )
> * if you quickly switch between applications, make sure that the latest application call is taken into account
> 
> in fact, you may need to update the apollo call inside the loadApplication method to use a takeUntil, based on the loadingApplication? so if you emit a new value while loading an app, you know you need to cancel the first apollo query, and only focus on the second one
> 
> hope that makes sense!

I think that the takeUntil with the loadingApplication emit would take care of those cases, I'll update the PR

unai-reliefapp:
> @unai-reliefapp
> 
> it looks quite better like that 👍
> 
> there may be a edge case, and I don't know how often it can happen, but we need to be sure that:
> 
> * if you exit an application ( back-office ), then you don't set applicationId in metadata ( so the app service doesn't contain app anymore )
> * especially if, from the back-office, you enter an app and quickly go back ( without letting time for the system to load the application )
> * if you quickly switch between applications, make sure that the latest application call is taken into account
> 
> in fact, you may need to update the apollo call inside the loadApplication method to use a takeUntil, based on the loadingApplication? so if you emit a new value while loading an app, you know you need to cancel the first apollo query, and only focus on the second one
> 
> hope that makes sense!

Actually the takeuntil is not needed, the leaveApplication method unsubscribes automatically from any ongoing application fetch
https://github.com/ReliefApplications/ems-frontend/blob/AB%23105244_activity_logs/libs/shared/src/lib/services/application/application.service.ts#L348
which is called on each loadapplication
https://github.com/ReliefApplications/ems-frontend/blob/AB%23105244_activity_logs/libs/shared/src/lib/services/application/application.service.ts#L246

AntoineRelief:
@unai-reliefapp 
wonderful!
I guess we had this issue before, so that's how we solved it
looks good then, please finish the PR and assign to me 👍

unai-reliefapp:
@AntoineRelief Explanation in the logger service as a comment for further references
also:
https://github.com/user-attachments/assets/b6a30500-1eeb-41f9-bb1e-0cbcffc6406e

PR #2764, opened 2025-02-18
AB#107812 Changes for File Upload Implementation

morgang-adappt:
# Description

Implement file uploads for email functionality. 

## Useful links

- Ticket: https://dev.azure.com/WHOHQ/EMSSAFE/_boards/board/t/HEMS%20-%20Email%20Notifications/Stories?workitem=107812
- Backend PR: https://github.com/ReliefApplications/ems-backend/pull/1191
- Function PR: https://dev.azure.com/WHOHQ/EMSSAFE/_git/emssafe-emailfunction/pullrequest/20212

## Type of change

- [x] This change requires a documentation update


libs/shared/src/lib/services/document-management/document-management.service.ts
```diff
@@ -34,46 +34,46 @@ interface PropertyQueryResponse {
  * Available properties from the CS API Documentation
  */
 export const CS_DOCUMENTS_PROPERTIES = [
-  { text: 'Aetiology', value: 'aetiologys', bodyKey: 'Aetiology' },
+  { text: 'Document Type', value: 'documenttypes', bodyKey: 'DocumentType' },
```
AntoineRelief: @morgang-adappt 
why did you change the order of properties there?

SafiMohammed: @AntoineRelief  We moved the mandatory and most relevant ones to the top

AntoineRelief: ok, thanks @SafiMohammed 
well, I'll revert that because previously they were ordered by alphabetical order which make more sense because mandatory & relevant only depend on the context of the upload

libs/shared/src/lib/components/email/components/email-attachment/email-attachment.component.ts
```diff
@@ -0,0 +1,359 @@
...
+      const uploadPromises = files.map(async (file) => {
+        const fileStream = await this.docManagement.transformFileToValidInput(
+          file
+        );
+        const url = `${this.environment.csApiUrl}/documents/drives/${driveId}/items`;
+
+        // If send as file
+        if (this.sendAsAttachment) {
+          this.bodyFilter = {
+            DocumentType: [50],
```
AntoineRelief: @SafiMohammed 
what are these value?
it would be good to indicate that in the associated ticket to help for the debugging

AntoineRelief: in addition, please note that the ids are not consistent between environments, so 50 doesn't mean the same in DEV & PROD

for example, in DEV it is "Temporary Attachment" but in PROD it is "Application Feedback" so we should find another way to do that

Vaishali-n: DocumentType:{
"id": 50,
"name": "Temporary Attachment"}

InformationConfidentiality:  {
"id": 2,
  "name": "Internal use only"
 }

I will update in the ticket as well @AntoineRelief

Vaishali-n: Please note if Jered wants to change it to something else, I will let you know

AntoineRelief: perhaps by adding that as environment variables as well

and I see most of the csDocUrl are missing in environment files so it would be good to add them
I'm pretty sure it hasn't been testing using the front-office, and was only tested using back-office, but end -users mostly use the front-office

morgang-adappt: Hey @AntoineRelief,

I'll add csDocUrl to the other environments now. Should we add this to web-widgets, as well as front-office? Is there a different URL we should use in prod and uat?

AntoineRelief: @morgang-adappt 
the only difference between the environment is the hems block

dev: hems-dev
uat: hems-test
prod: hems

and yes all environments files ( default / azure.dev / azure.uat / azure.prod ) of widgets / front-office / back-office should have the correct value

morgang-adappt: @AntoineRelief 
This is pushed

AntoineRelief:
@morgang-adappt @SafiMohammed 

these are not mandatory parameters:

Capture d’écran 2025-02-18 à 16 26 56

Document Type may be, but I'm not 100% sure, but Confidentiality isn't mandatory

For example, for SA, we upload files with IMS Role set to Surveillance and it automatically set the Confidentialy

libs/shared/src/lib/components/email/steps/preview/preview.component.ts
```diff
@@ -461,6 +468,35 @@ export class PreviewComponent
           if (this.subjectHtmlRef?.nativeElement) {
             this.subjectHtmlRef.nativeElement.innerHTML = this.subjectString;
           }
+          if (this.attachmentHtmlRef?.nativeElement) {
+            const attachmentUrls =
+              this.emailService.finalEmailPreview.attachments.files;
+
+            if (Array.isArray(attachmentUrls) && attachmentUrls.length > 0) {
+              const attachmentLinks = attachmentUrls
```
AntoineRelief: @morgang-adappt @SafiMohammed 

I think this is a duplication of code because the email function app should certainly be responsible from building the links, not the front-end
as the function app is capable of sending these links without relying on the front-end to pass the updated values after previewing the email

libs/shared/src/lib/components/email/constant.ts
```diff
@@ -26,3 +26,20 @@ export const commonServiceFields = [
   { key: 'LocationType', label: 'LocationType' },
   { key: 'InternalExternal', label: 'InternalExternal' },
 ];
+
+/**
+ * proxy Graphql Path
+ */
+export const proxyGraphqlPath = 'CS_DEV/graphql/';
```
AntoineRelief: @morgang-adappt @SafiMohammed 

that works for DEV, not for UAT, not for PROD

it's being used in several places, and it prevents any deployment in any other environment than DEV

see how it's being used in document-management service
there is one method called:
createCSApolloClient
that creates the client and enable to directly query the CS GQL API
we also have the common-services API, which does something similar

in addition, you're duplicating the code as some queries ( like the ones to fetch the possible IDs for the CS properties ) are quite similar to what we already declared in document management so instead we should reuse the code or refactor it a bit so it can work fine

the email-attachment component should rely on existing services I mentionned previously for the query, and not rely on API proxy

libs/shared/src/lib/components/email/steps/preview/preview.component.ts
AntoineRelief: in addition you should use something like promise.all because right now it's doing the requests one by one which can take some time to resolve

libs/shared/src/lib/components/email/components/email-attachment/email-attachment.component.html
```diff
@@ -0,0 +1,123 @@
...
+          let i = index
+        "
+        [removable]="true"
+        (removed)="removeFile(i)"
+      >
+        {{ file.fileName }}
+      </ui-chip>
+    </div>
+    <div>
+      <p class="noteStyle">Please note: You can upload a maximum of 5 files, with each file not exceeding 7MB in size.</p>
```
AntoineRelief: @morgang-adappt @SafiMohammed 
please avoid direct html text in Angular templates

AntoineRelief:
@morgang-adappt @SafiMohammed 

just to be sure
I still need to run the back-end with the changes you did previously right?

morgang-adappt:
> @morgang-adappt @SafiMohammed
> 
> just to be sure I still need to run the back-end with the changes you did previously right?

@AntoineRelief 
Yes, the backend changes are still required.

AntoineRelief:
@morgang-adappt @SafiMohammed 

FYI

I deactivated the logic that was fetching the possible values for the doc properties for attach as file as it was deactivated
it was just doing unneeded API calls at this stage

In addition, I found an issue with the upload you can see in commit:
[96f5809](https://github.com/ReliefApplications/ems-frontend/pull/2764/commits/96f58090cd28ff8db850dabd6fc655a7664535eb)
if the user would drop a file, then open again the file selector and cancel, it would break navigation

PR #2768, opened 2025-02-26
feat/AB-DL-CRUD Implementation 

morgang-adappt:
# Description

## New Feat: DL CRUD
* Quick send distribution list selection is now restricted to select distribution lists that do not have the grid's resource configured as the resource in select with filter. 
* Distribution lists can now be previewed from the distribution list tab in email notifications in a reusable component. 
### Create feature:
* Add New Distribution List should be present to allow user to create a new distribution list.
* Create feature should have all the options (add manually, select with filter, select from common services, use combination) that are available in the DL page of the email notification
* Any newly create DL from the DL list page should be visible in the DL page of the email notification and in the Quick send DL dropdown
### Edit feature:
* Each DL in the list should be provided with the edit and delete buttons.
* Any modification made to the DL should be reflected in the EN and in the quick send which uses the same DL
### Delete feature:
* On deleting any DL, the following confirmation message should appear "Please be aware that deleting this Distribution List (DL) will impact all instances where it is currently in use. Do you want to proceed with the deletion?"
* On confirming the deletion, that DL instance should be removed from EN and from Quick send
* For the EN associated to the deleted DL, the send email icon should be greyed out, thus forcing the user to edit the EN which lands them on the first stepper. This allows the user to navigate to DL page and add a new DL or use existing ones

## Fix: File attachments are now deleted from the document management system when removed by the user when configuring an email notification. 
* A delete request is made to the CS API to delete a file when it is removed by a user. 

## Improvement: Time stamps can now be inserted into the body
* Following the function changes to replace date macros in the body, the frontend now has a dropdown to insert timestamps into the body. 

# Useful links

- Backend PR:  https://github.com/ReliefApplications/ems-backend/pull/1193
- Function PR: https://dev.azure.com/WHOHQ/EMSSAFE/_git/emssafe-emailfunction/pullrequest/20329

# Type of change


AntoineRelief:
@morgang-adappt 
I really prefer you create distinct PRs when the commits are not related
Based on what I see there, it could have been 3 PRs
it takes more time to check one PR with different work items in it than 3 PRs with clear distinction of purpose

for this one, it's already created so I'll check it but in the future please keep that in mind, thanks!

morgang-adappt:
> @morgang-adappt I really prefer you create distinct PRs when the commits are not related Based on what I see there, it could have been 3 PRs it takes more time to check one PR with different work items in it than 3 PRs with clear distinction of purpose
> 
> for this one, it's already created so I'll check it but in the future please keep that in mind, thanks!

Hey @AntoineRelief,

Thank you for the feedback, I agree with you on this point, and I'm sorry for the PR being so large. Going forward, we will get the PRs opened in a cleaner, more atomic manner. 

Additionally, our QA has spotted a few issues with resource selection in this PR. Once these fixes are signed off, I'll update the PR ASAP, if that works for you?

AntoineRelief:
hey @morgang-adappt 

are you done with the changes on CRUD?
I can see you pushed new commits in the back-end branch a few minutes ago

just to know if I can start reviewing the PRs or if I should wait for more changes

btw, i'll change the target branches to "beta", just in case the changes aren't validated before next saturday, so we keep "next" limited to validated changes, and if we need to go for a hotfix after release we can start from next branch

morgang-adappt:
@AntoineRelief 

The backend push was just to tidy a comment, the code itself is unchanged. 

The frontend CRUD work appears to have a regression which breaks resource selection in some distribution list configuration scenarios. 
image

The function and backend PRs are ready to review

AntoineRelief:
> @AntoineRelief
> 
> The backend push was just to tidy a comment, the code itself is unchanged.
> 
> The frontend CRUD work appears to have a regression which breaks resource selection in some distribution list configuration scenarios. image
> 
> The function and backend PRs are ready to review

thanks @morgang-adappt 

in that case, I'll wait for this regression to be fixed to review the CRUD, and I'll start by reviewing the other elements of work that were included in these PRs

morgang-adappt:
Hey @AntoineRelief,

The PR has been updated with the fixes mentioned.

libs/shared/src/i18n/en.json
```diff
@@ -174,8 +174,14 @@
 		"dismiss": "Dismiss",
 		"display": "Display",
 		"distributionList": {
+			"DLCreated": "Distribution list created",
+			"DLDeleted": "Distribution list deleted successfully",
+			"DLEdited": "Distribution list edited successfully",
+			"DeleteAlert": "Please be aware that deleting this Distribution List (DL) will impact all instances where it is currently in use. Do you want to proceed with the deletion?",
```
AntoineRelief: @morgang-adappt 
can I do quick updates on the i18n values to make it more consistent?
I don't know if any text was agreed, so just to be sure

libs/shared/src/lib/services/application/graphql/queries.ts
```diff
@@ -122,3 +122,12 @@ export const GET_APPLICATION_BY_ID = gql`
     }
   }
 `;
+
+/** Graphql request for adding a template of an application */
+export const DELETE_EMAIL_DISTRIBUTION_LIST = gql`
```
AntoineRelief: @NehaSakariya-Adappt 

FYI only
please put the queries at the same level than the elements that use them
it's not being used by the application service so it should be declared there, but in the graphql folder for the email service

AntoineRelief: & there is a distinction between queries & mutations

AntoineRelief: also, the comment is incorrect, please double check the comments after copy pasting

libs/shared/src/i18n/en.json
morgang-adappt: @AntoineRelief 
Feel free to update, thank you!

libs/shared/src/lib/services/application/graphql/queries.ts
morgang-adappt: Noted, I'll pass this on to the rest of the team. Thank you for the refactor on this!

libs/shared/src/lib/components/email/components/create-distribution/create-distribution.component.html
```diff
@@ -0,0 +1,25 @@
+<div class="flex flex-col items-center">
+  <div class="max-w-5xl w-full">
+    <app-select-distribution
+      [DL_QuickAction]="DL_QuickAction"
```
AntoineRelief: @NehaSakariya-Adappt 
FYI this naming is "accepted" by the code, but is invalid in terms of code quality
you should avoid using _ in variables as well ad starting a variable name with capitalized letters

cc @morgang-adappt

morgang-adappt: Hey @AntoineRelief,

Neha's left for the day, I'm happy to refactor this if it would help?

AntoineRelief: thanks @morgang-adappt 
I'll do any refactor change if needed, to avoid any conflict

AntoineRelief:
@morgang-adappt @NehaSakariya-Adappt 

just found a critical bug with the PR

I'll share the video in the function app PR to avoid putting it in a github repo

morgang-adappt:
> @morgang-adappt @NehaSakariya-Adappt
> 
> just found a critical bug with the PR
> 
> I'll share the video in the function app PR to avoid putting it in a github repo

Thanks for bringing this to our attention, I've passed the details onto the team and have started looking into this.

AntoineRelief:
@morgang-adappt @NehaSakariya-Adappt 

some feedback regarding possible refactor if you have time to check that, for better integration with the rest of the code

**email service:**
- `isDLEdit` is not a valid variable name as it lacks clear meaning, it should be something like isEditingDL for example, or, isEditingDistributionList ( sometimes, longer names can make the code easier to understand because DL is not straightforward )
- `DL_Data` is not a valid variable name, you could use distributionList instead. JavaScript and TypeScript typically use camelCase or PascalCase, so DL_Data looks odd
- same comment for `cs_detail` ( around line 718 )
- same comment for `set_Filter_CS_Value`

**layout component:**
- clear the constructor as some useless services injections were added in this PR

**select distribution component**
- same comment as above about: `DL_QuickAction` ( which I found in many places )

**resource-select**
- I haven't been able to check in details why you add the `quickActionResource` property but I'm pretty sure you don't need it as it seems quite odd to me than many elements in the code use this resource select component without needing this property. I believe the real issue is about how you set the form control value or the selectedElements there:

<shared-resource-select
          formControlName="resource"
          [selectedElements]="[resource]"
          [quickActionResource]="
            DL_QuickAction?.resource ? selectedResourceName : ''
          "
        ></shared-resource-select>

morgang-adappt:
> > @morgang-adappt @NehaSakariya-Adappt
> > just found a critical bug with the PR
> > I'll share the video in the function app PR to avoid putting it in a github repo
> 
> Thanks for bringing this to our attention, I've passed the details onto the team and have started looking into this.

Hey @AntoineRelief 

We've been trying to reproduce this for the last hour and are struggling, we've been testing in both localhost connected to a local DB instance and our POC environment. Would it be possible to schedule a call now to debug this further?

NehaSakariya-Adappt:
HI @AntoineRelief 

For **resource-select**
In the Resource-select.component.ts file, the quickActionResource variable is used to 
send a specific resource name when previewing the distribution list from the grid data. 
(The resource name assigned to **quickActionResource** variable will match the name of the resource displayed in the grid.)

AntoineRelief:
@morgang-adappt @NehaSakariya-Adappt 

I think I found what would cause the issue
I'll ping you on the specific method

libs/shared/src/lib/components/email/steps/layout/layout.component.ts
```diff
@@ -140,13 +142,17 @@ export class LayoutComponent
...
+   * @param http - The http client for making HTTP requests.
    */
   constructor(
     private fb: FormBuilder,
     private editorService: EditorService,
     public emailService: EmailService,
     public snackbar: SnackbarService,
-    public translate: TranslateService
+    public translate: TranslateService,
+    private restService: RestService,
```
AntoineRelief: @morgang-adappt @NehaSakariya-Adappt 
you should choose between http & restService, and preferably use the restService which can do exactly what the http does
I think it's not used there so you should remove it, but in several files there is this duplication

libs/shared/src/lib/components/email/steps/layout/layout.component.ts
```diff
@@ -266,10 +272,107 @@ export class LayoutComponent
...
+      this.loadDistributionList();
     }
-    this.getBlockData();
+  }
+
+  /**
+   * Loads the distribution list.
+   *
+   */
+  async loadDistributionList() {
```
AntoineRelief, line 287: @morgang-adappt @NehaSakariya-Adappt 
this is the method where I found an issue

using distribution lists that were generated before this PR was created, which is exactly what we have for signal apps

how to replicate:
- open an app with current state of code in UAT ( next branches for back & front + alpha for email )
- create a grid
- create a DL from that grid ( only allows input emails )
- save the grid & test it, the DL is working

Now:
- open the same app but with the PRs
- open the same grid
- try to trigger the same action
- the DL is blank

there is a lot of tests in that method and I guess one of them is failling because if I print the query, I get the exact same DL I created, so it's not a data issue, but it's likely to be a check somewhere there, as the email function is not triggered in that case

btw, the email service is really getting bigger and bigger with all the latest PRs and it's becoming quite hard to debug it because of all the places it's being used
it's just a suggestion, but the problem I see with this service is that it controls the state of too many components, instead of only being responsible from creating the forms, sending the requests
the state of components would be better managed if you would implement it at component level, and that would certainly avoid a lot of issues and simplify the updates

AntoineRelief:
> HI @AntoineRelief
> 
> For **resource-select** In the Resource-select.component.ts file, the quickActionResource variable is used to send a specific resource name when previewing the distribution list from the grid data. (The resource name assigned to **quickActionResource** variable will match the name of the resource displayed in the grid.)

hey @NehaSakariya-Adappt 
I understand that, but what I want to highlight is that you're editing a core component, being used by about 15 other components, and you add an input to it that is only used by one of these 15 components, and with a naming that isn't understandable out of the context of that component

also, I don't understand why we would limit the resource in that situation, I think that's quite a limitation in fact, because we can easily imagine use cases where users would build distribution lists based on other resources
Where is that requirement coming from?

Vaishali-n:
> > HI @AntoineRelief
> > For **resource-select** In the Resource-select.component.ts file, the quickActionResource variable is used to send a specific resource name when previewing the distribution list from the grid data. (The resource name assigned to **quickActionResource** variable will match the name of the resource displayed in the grid.)
> 
> hey @NehaSakariya-Adappt I understand that, but what I want to highlight is that you're editing a core component, being used by about 15 other components, and you add an input to it that is only used by one of these 15 components, and with a naming that isn't understandable out of the context of that component
> 
> also, I don't understand why we would limit the resource in that situation, I think that's quite a limitation in fact, because we can easily imagine use cases where users would build distribution lists based on other resources Where is that requirement coming from?

@AntoineRelief 

This requirement is logical for Email Notifications, where an admin or back office user with access to all resources configures and sends the email. However, for Quick Send from Grid, the email can be triggered by an end user who doesn't have access to other resources. Therefore, exposing other resources or related data to them is not feasible. Let me know if this makes sense

AntoineRelief:
> > > HI @AntoineRelief
> > > For **resource-select** In the Resource-select.component.ts file, the quickActionResource variable is used to send a specific resource name when previewing the distribution list from the grid data. (The resource name assigned to **quickActionResource** variable will match the name of the resource displayed in the grid.)
> > 
> > 
> > hey @NehaSakariya-Adappt I understand that, but what I want to highlight is that you're editing a core component, being used by about 15 other components, and you add an input to it that is only used by one of these 15 components, and with a naming that isn't understandable out of the context of that component
> > also, I don't understand why we would limit the resource in that situation, I think that's quite a limitation in fact, because we can easily imagine use cases where users would build distribution lists based on other resources Where is that requirement coming from?
> 
> @AntoineRelief
> 
> This requirement is logical for Email Notifications, where an admin or back office user with access to all resources configures and sends the email. However, for Quick Send from Grid, the email can be triggered by an end user who doesn't have access to other resources. Therefore, exposing other resources or related data to them is not feasible. Let me know if this makes sense

@Vaishali-n 
on top of it there is a permission system that would hide the resources if the user cannot see them
and still, I'm pretty sure some use cases could exist where users would need to send emails from a grid using another resource they have access to

Vaishali-n:
> > > > HI @AntoineRelief
> > > > For **resource-select** In the Resource-select.component.ts file, the quickActionResource variable is used to send a specific resource name when previewing the distribution list from the grid data. (The resource name assigned to **quickActionResource** variable will match the name of the resource displayed in the grid.)
> > > 
> > > 
> > > hey @NehaSakariya-Adappt I understand that, but what I want to highlight is that you're editing a core component, being used by about 15 other components, and you add an input to it that is only used by one of these 15 components, and with a naming that isn't understandable out of the context of that component
> > > also, I don't understand why we would limit the resource in that situation, I think that's quite a limitation in fact, because we can easily imagine use cases where users would build distribution lists based on other resources Where is that requirement coming from?
> > 
> > 
> > @AntoineRelief
> > This requirement is logical for Email Notifications, where an admin or back office user with access to all resources configures and sends the email. However, for Quick Send from Grid, the email can be triggered by an end user who doesn't have access to other resources. Therefore, exposing other resources or related data to them is not feasible. Let me know if this makes sense
> 
> @Vaishali-n on top of it there is a permission system that would hide the resources if the user cannot see them and still, I'm pretty sure some use cases could exist where users would need to send emails from a grid using another resource they have access to

@AntoineRelief If the permissions don't allow users to see the resources or data, configuring it is pointless since the user won't be able to access it. If admin or back-office users with access to all resources want to configure the distribution list (DL) from different resources, they should use Email Notification instead of Quick Send.

AntoineRelief:
> > > > > HI @AntoineRelief
> > > > > For **resource-select** In the Resource-select.component.ts file, the quickActionResource variable is used to send a specific resource name when previewing the distribution list from the grid data. (The resource name assigned to **quickActionResource** variable will match the name of the resource displayed in the grid.)
> > > > 
> > > > 
> > > > hey @NehaSakariya-Adappt I understand that, but what I want to highlight is that you're editing a core component, being used by about 15 other components, and you add an input to it that is only used by one of these 15 components, and with a naming that isn't understandable out of the context of that component
> > > > also, I don't understand why we would limit the resource in that situation, I think that's quite a limitation in fact, because we can easily imagine use cases where users would build distribution lists based on other resources Where is that requirement coming from?
> > > 
> > > 
> > > @AntoineRelief
> > > This requirement is logical for Email Notifications, where an admin or back office user with access to all resources configures and sends the email. However, for Quick Send from Grid, the email can be triggered by an end user who doesn't have access to other resources. Therefore, exposing other resources or related data to them is not feasible. Let me know if this makes sense
> > 
> > 
> > @Vaishali-n on top of it there is a permission system that would hide the resources if the user cannot see them and still, I'm pretty sure some use cases could exist where users would need to send emails from a grid using another resource they have access to
> 
> @AntoineRelief If the permissions don't allow users to see the resources or data, configuring it is pointless since the user won't be able to access it. If admin or back-office users with access to all resources want to configure the distribution list (DL) from different resources, they should use Email Notification instead of Quick Send.

@Vaishali-n 
they will only see the resources they have access to, so they can still see some resources in that part
honestly I find it rather strange from a user perspective that would have access to both the quick send & distribution list builder to find two different behaviors for the same thing
But in comments above I was wondering if this is a requirement coming from the client

in that case okay we can go with it, but the code can be improved, and you could for example auto select the resource instead of having the click on the resource selector, to select it. In addition, enabling filtering is useless if you only see one option in the list
so you could simply have the select with the resource automatically selected, and just make sure that if there is no fields selected, then it doesn't do anything at the end
for the user experience that would require less clicks, and it would certainly be more usable

Vaishali-n:
> > > > > > HI @AntoineRelief
> > > > > > For **resource-select** In the Resource-select.component.ts file, the quickActionResource variable is used to send a specific resource name when previewing the distribution list from the grid data. (The resource name assigned to **quickActionResource** variable will match the name of the resource displayed in the grid.)
> > > > > 
> > > > > 
> > > > > hey @NehaSakariya-Adappt I understand that, but what I want to highlight is that you're editing a core component, being used by about 15 other components, and you add an input to it that is only used by one of these 15 components, and with a naming that isn't understandable out of the context of that component
> > > > > also, I don't understand why we would limit the resource in that situation, I think that's quite a limitation in fact, because we can easily imagine use cases where users would build distribution lists based on other resources Where is that requirement coming from?
> > > > 
> > > > 
> > > > @AntoineRelief
> > > > This requirement is logical for Email Notifications, where an admin or back office user with access to all resources configures and sends the email. However, for Quick Send from Grid, the email can be triggered by an end user who doesn't have access to other resources. Therefore, exposing other resources or related data to them is not feasible. Let me know if this makes sense
> > > 
> > > 
> > > @Vaishali-n on top of it there is a permission system that would hide the resources if the user cannot see them and still, I'm pretty sure some use cases could exist where users would need to send emails from a grid using another resource they have access to
> > 
> > 
> > @AntoineRelief If the permissions don't allow users to see the resources or data, configuring it is pointless since the user won't be able to access it. If admin or back-office users with access to all resources want to configure the distribution list (DL) from different resources, they should use Email Notification instead of Quick Send.
> 
> @Vaishali-n they will only see the resources they have access to, so they can still see some resources in that part honestly I find it rather strange from a user perspective that would have access to both the quick send & distribution list builder to find two different behaviors for the same thing But in comments above I was wondering if this is a requirement coming from the client
> 
> in that case okay we can go with it, but the code can be improved, and you could for example auto select the resource instead of having the click on the resource selector, to select it. In addition, enabling filtering is useless if you only see one option in the list so you could simply have the select with the resource automatically selected, and just make sure that if there is no fields selected, then it doesn't do anything at the end for the user experience that would require less clicks, and it would certainly be more usable

@AntoineRelief Regarding the requirement, I will confirm with Jered and let you know. As for the UX, I agree it can be improved. The resource should be selected by default instead of providing a dropdown

morgang-adappt:
Hey @Matthis-M-ReliefApps and @AntoineRelief ,
 
We'll be updating the PRs today with the fixes, refactors, and improvements discussed.
We have not been able to refactor the state from the email service to the component level following Antoine's suggestion, but will take this up in the future.
 
- Distribution Lists configured on the previous code base in UAT now work in quick send.
- Variables have been renamed for clarity.
- Unnecassary fields have been removed the layout constructor.
- `quickActionResource` has been removed from the `resource-select` component.
 
Additionally, we found a few regressions, which have been fixed:
 
- In the send seperate email flow, the send email button was erroneously disabled.
  - The send button is now disabled in cases where the distribution list associated with a send seperate email has been deleted.
- Frontend permissions checks have been implemented for edition and deletion of distribution lists in the distribution list tab
- Distribution list selection in quick send would filter out all distribution lists configured with a resource filter, the desired behaviour is to filter out distribution lists whose resource does not match the grid resource.
- Distribution lists being edited would clear the resource selection when configured with use combination.
 
On the backend, I have made one small change. Our `deleteEmailDistributionList` mutation now checks for update permission on the `EmailNotification` subject, rather than delete on the legacy `DistributionList` subject, because this involves updating EmailNotifications and is in line with the permission on the frontend.
 
On the function, there is a fix for fields of type `resources` not displaying their label correctly, and the function app now checks that some fields have been selected before attempting to fetch resource data.

morgang-adappt:
Hey @Matthis-M-ReliefApps and @AntoineRelief 

The PRs have been updated with the fixes discussed.

HasnatL-Adapptlimited:
Hey @AntoineRelief @Matthis-M-ReliefApps 
Added one more commit on top of this for a bug QA noticed - When user tries to send email from  EN with common services DL, the send email button was disabled and DL was not visible in preview screen.

CC: @morgang-adappt

PR #2806, opened 2025-08-28
AB#118658 - ABC Duplicate record

Joselgc1:
# Description

Enable records to be duplicated by showing a button while seeing record details

## Useful links

https://github.com/ReliefApplications/ems-frontend/tree/AB%23118658

## Type of change

### DIFF
--- a/libs/shared/src/lib/components/record-modal/record-modal.component.html
+++ b/libs/shared/src/lib/components/record-modal/record-modal.component.html
@@ -12,6 +12,15 @@
           [survey]="survey"
           [surveyNext]="surveyNext || undefined"
         ></shared-form-actions>
+
+        <ui-button
+          *ngIf="form?.canCreateRecords && !data.compareTo"
+          variant="primary"
+          category="secondary"
+          (click)="onDuplicate()"
+        >
+          Duplicate record
+        </ui-button>
       </div>
       <ui-tabs
         [selectedIndex]="(selectedPageIndex$ | async)!"
--- a/libs/shared/src/lib/components/record-modal/record-modal.component.html
+++ b/libs/shared/src/lib/components/record-modal/record-modal.component.html
@@ -19,7 +19,7 @@
           category="secondary"
           (click)="onDuplicate()"
         >
-          Duplicate record
+          {{ 'common.duplicateRecord' | translate }}
         </ui-button>
       </div>
       <ui-tabs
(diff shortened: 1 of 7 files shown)


libs/shared/src/lib/components/record-modal/record-modal.component.html
```diff
@@ -12,6 +12,15 @@
...
           [surveyNext]="surveyNext || undefined"
         ></shared-form-actions>
+
+        <ui-button
+          *ngIf="form?.canCreateRecords && !data.compareTo"
+          variant="primary"
+          category="secondary"
+          (click)="onDuplicate()"
+        >
+          Duplicate record
```
AntoineRelief: @Joselgc1 
Quick change, you should add it as a key in en.json & fr.json ( use text in fr: "Dupliquer l'enregistrement" ) & load it there using translate pipe

libs/shared/src/lib/components/record-modal/record-modal.component.ts
```diff
@@ -263,6 +263,30 @@ export class RecordModalComponent
...
+      data: {
+        template: this.form.id,
+        prefillData: this.record?.data,
+        askForConfirm: false,
+      },
+      autoFocus: false,
+    });
+    dialogRef.closed.pipe(takeUntil(this.destroy$)).subscribe((value: any) => {
+      if (value) {
+        this.dialogRef.close(value);
```
AntoineRelief: @Joselgc1 
in case this is automatically closing the modal when a new record is saved, you can remove it
user will continue to view the modal that initiated the duplication

libs/shared/src/lib/components/ui/core-grid/core-grid.component.ts
```diff
@@ -1163,7 +1163,9 @@ export class CoreGridComponent
       dialogRef.closed
         .pipe(takeUntil(this.destroy$))
         .subscribe((value: any) => {
-          if (value) {
+          if (value?.data) {
```
AntoineRelief: @Joselgc1 
not sure what it does?

Joselgc1: I added this so that once the new duplicated record is created, the modal was closed and the dashboard was automatically refreshed with the new record row shown. However, it doesn't work if we keep the original record modal open so I'll remove it

PR #2825, opened 2025-10-06
AB#118658 - Record cloning improvements

Joselgc1:
# Description

Users are able to choose which page the user is navigated to after the user Saves the new record that's based on a clone. It implements the already existing Navigate to Page and Target URL functions.

## Type of change

# Checklist:

( * == Mandatory ) 


libs/shared/src/lib/components/action-button/action-button.component.ts
```diff
@@ -419,8 +435,44 @@ export class ActionButtonComponent
...
+                  if (nav.targetPage?.enabled && nav.targetPage.pageUrl) {
+                    let fullUrl = this.getPageUrl(
+                      nav.targetPage.pageUrl as string
+                    );
+                    if (nav.targetPage.field && value?.data) {
+                      const fieldPath = nav.targetPage.field;
+                      const paramValue = get(value.data, fieldPath);
+                      fullUrl = `${fullUrl}?${fieldPath}=${paramValue}`;
+                    }
+                    if (fullUrl.startsWith('./'))
```
AntoineRelief: @Joselgc1 
do you remember why you did that?

Joselgc1: I thought it was a mistake since previously the fieldPath was always set to id and the user is allowed to select many other parameters to query

AntoineRelief: okay! thanks for the quick answer @Joselgc1 , I'll test without it

libs/shared/src/lib/components/action-button/action-button.component.ts
```diff
@@ -342,6 +357,7 @@ export class ActionButtonComponent
           disableClose: true,
           data: {
             ...(this.actionButton.editRecord && { recordId: this.contextId }), // button must be hidden in html if editRecord is enabled & no contextId
+            ...(this.actionButton.cloneRecord && { recordId: this.contextId }), // button must be hidden in html if cloneRecord is enabled & no contextId
```
AntoineRelief: @Joselgc1 
not so critical because I could easily fix it, and I see this comment was misleading, but by adding this line, in fact we're not cloning anymore but editing the record we open 😅

AntoineRelief:
@Joselgc1 

really good work on this one 👍

only issue I saw is about the cloning that wasn't cloning anymore but editing, but I think the comment that was left was misleading and confused you

what I did:
- I simplified the logic, because the "enabled" booleans are in fact not needed and can be calculated on the go based on the data we store ( update in both front-end & back-end )

what I'll do:
- test if it's possible to redirect to a form page & load a record if id is in url (out of the scope, based on the requirements I gave you )
- Add possibility to duplicate action buttons ( out of scope as well )

Thanks José!

apps/back-office/src/app/components/edit-action-button-modal/edit-action-button-modal.component.ts
```diff
@@ -182,13 +193,16 @@ export class EditActionButtonModalComponent
           })
           .subscribe({
             next: ({ data }) => {
+              const queryTemp: any = data.resource;
```
AntoineRelief: @Joselgc1 
(Just for your information, I'll work on it)
I think you copy pasted the code from elsewhere in the code (one part of the code that wasn't written by us)
you can eliminate about 2 lines there, just by writting:

this.availableFields = this.queryBuilder.getFields(data.resource.queryName!);

also, there is a repetition, because the sendNotificationFields has the same value

as it's one of the only places I can give some feedback on, as the rest is super, I'm a bit picky ;)

so, from this, keep in mind:
- avoid any when possible
- eliminate unnecessary code lines ( sometimes, it's better to keep it more structured so it's more readable, but I don't think it applies there)

PR #2833, opened 2025-10-31
AB#68534 - People Picker for forms

Joselgc1:
# Description

So Far:

The new People field was added to SurveyJS and is fully functional. It appears in the toolbox, with a configurable placeholder, and connects to Common Services via the existing csClient Apollo setup. The dropdown supports lazy loading with a fixed debounce time, and has configurable minimum character requirement and page size parameters for fetching results. Options are displayed in the “First Last (email)” format, and the selected value correctly stores the Azure userid. Pagination is also implemented as the user scrolls.

Remaining Work:

All the Grid integrations are still pending: People field should load only visible records in grids and all records during export. Additionally, storing basic user details (first name, last name, email) alongside the userid in the metadata to support more efficient sorting and filtering.

## Type of change

# Checklist:

( * == Mandatory ) 


AntoineRelief:
@Joselgc1 

This will be the main priority for you on the project

Here are the changes I'd like, compared to the original scope (or already in original scope, but good to mention them again):
- people picker can be single or multi select. Meaning we need 2 question types, that can easily be identified, like people-dropdown & people-tagbox
- Same rules than before for selection / searching etc ( in form edition so, would it be tagbox or dropdown )
- for grids, we will display users as, if it's a tagbox: FirstName LastName (email1), FirstName LastName (email1)
- if it's a dropdown: FirstName LastName (email1)
- Keep same display when displaying data in summary cards
- If possible, also allow aggregations to work fine with the data ( in that case, aggregated data can also use people data like: FirstName LastName (email1)

Joselgc1:
The People Picker feature is now fully completed. Building on the initial version, which already had the basic field working in SurveyJS with lazy-loaded search and correct value storage, we now support two field types: People (Single) and People (Multiple), both available in the SurveyJS toolbox. Search has been switched from the old Apollo client setup to the Common Services REST endpoint, still using debounce, and showing results in the "FirstName LastName (email)" format. The multi-select field now displays chosen users as removable chips for a much clearer UX.

We improved how values are stored. Instead of saving just the userid, each selection now stores the full user object (firstname, lastname, email, userid). This solves the remaining Grid-related tasks: grids can display names immediately without extra lookups, and sorting and filtering work smoothly because all relevant metadata is already available. Grid rendering is now fully functional for both single- and multi-select fields.

Pagination is no longer needed since the REST endpoint only supports whole-word matching.

All pending work mentioned in the first description has been addressed: the fields work inside grids, exports include complete user info, and the metadata structure is now optimized for both performance and usability.

PR #2838, opened 2026-01-06
AB#121666

AntoineRelief:
# Description

Implement per-row visibility for custom action buttons in grid widgets: each button now supports a [filter](vscode-file://vscode-app/opt/visual-studio-code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) that determines, on a record-by-record basis, whether the button is shown, and the action column width is computed from the number of buttons and label length instead of being hardcoded.

## Type of change

# How Has This Been Tested?

- Open a grid widget settings and create 2+ custom row actions with the same columnLabel and different filters.
- Verify that on each row, only the buttons whose filter matches are displayed.
- Create buttons with different columnLabel values and verify one column per label is rendered.
- Verify the column width adjusts to the number of buttons and label length.

## Screenshots

image

image

# Checklist:

( * == Mandatory ) 

- [x] * I have put the ticket for review, adding the oort-backend team to the list of reviewers

# More explanation


libs/shared/src/lib/components/ui/core-grid/grid/grid.component.ts
```diff
@@ -455,6 +455,67 @@ export class GridComponent
...
+   * @param group The action group column definition.
+   * @returns The action buttons to render for this row / group.
+   */
+  public getRowActions(
+    dataItem: any,
+    group: { label: string; actions: ActionButton[] }
+  ): ActionButton[] {
+    const rowGroups: { label: string; actions: ActionButton[] }[] =
+      dataItem?.actions ?? [];
+    // If the backend did not populate per-row actions, fall back to all actions.
```
AntoineRelief: @GhaziAlibi 
Only mistake I could find
If the back-end doesn't return any action, then it means they all got filtered out, and so, we shouldn't show them

PR #2845, opened 2026-03-30
AB#128603 - ABC - Custom label for form buttons

Joselgc1:
# Description

Forms were always using the same save and draft labels, including when the same form was opened inline or in a modal. Admins can now configure normal text or SurveyJS-style expressions for both actions in the Form Builder, with separate modal overrides available under Navigation > Advanced so labels like “Save and Send” or “Update draft” can change based on answers such as status, while still falling back to the existing translated defaults when no custom value is set.

## Type of change

# Screenshots

image

image

To change a form's labels

image

image

image

To change a form's modal Labels

image

image

# Checklist:

( * == Mandatory ) 

### DIFF
--- a/libs/shared/src/lib/components/form-modal/form-modal.component.html
+++ b/libs/shared/src/lib/components/form-modal/form-modal.component.html
@@ -71,9 +71,9 @@
         {{ 'common.next' | translate }}
       </ui-button>
     </ng-container>
-    <!-- <ui-button [disabled]="disableSaveAsDraft" (click)="saveAsDraft()">{{
-      'components.form.draftRecords.save' | translate
-    }}</ui-button> -->
+    <ui-button [disabled]="disableSaveAsDraft" (click)="saveAsDraft()">{{
+      saveAsDraftButtonLabel || ('components.form.draftRecords.save' | translate)
+    }}</ui-button>
     <ui-button
       category="secondary"
       variant="primary"
@@ -81,7 +81,7 @@
       (click)="submit()"
       cdkFocusInitial
     >
-      {{ 'components.record.modal.update' | translate }}
+      {{ saveButtonLabel || ('components.record.modal.update' | translate) }}
     </ui-button>
   </ng-container>
 </ui-dialog>
--- a/libs/shared/src/lib/components/form-modal/form-modal.component.html
+++ b/libs/shared/src/lib/components/form-modal/form-modal.component.html
@@ -72,7 +72,7 @@
       </ui-button>
     </ng-container>
     <ui-button [disabled]="disableSaveAsDraft" (click)="saveAsDraft()">{{
-      saveAsDraftButtonLabel || ('components.form.draftRecords.save' | translate)
+      modalSaveAsDraftButtonLabel || ('components.form.draftRecords.save' | translate)
     }}</ui-button>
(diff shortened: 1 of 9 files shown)


libs/shared/src/lib/components/form-modal/form-modal.component.html
```diff
@@ -71,17 +71,17 @@
...
+    }}</ui-button>
     <ui-button
       category="secondary"
       variant="primary"
       [disabled]="saving"
       (click)="submit()"
       cdkFocusInitial
     >
-      {{ 'components.record.modal.update' | translate }}
+      {{ saveButtonLabel || ('components.record.modal.update' | translate) }}
```
AntoineRelief: @Joselgc1 
Requested change
The save button label should not be the same between the form modal & the form
By default, the text isn't the same, so we must have 2 distinct options

libs/shared/src/lib/utils/survey-form-action-labels.util.ts
```diff
@@ -0,0 +1,53 @@
...
+export const SURVEY_PROP_SAVE_AS_DRAFT_BUTTON_LABEL = 'saveAsDraftButtonLabel';
+
+/**
+ * Evaluates a single form action button label from survey-level expression settings.
+ *
+ * @param survey Active survey model
+ * @param propertyName Serializer property name (e.g. saveButtonLabel)
+ * @returns Evaluated label, or empty string if unset (template uses default i18n)
+ */
+export function evaluateSurveyActionButtonLabel(
```
AntoineRelief, line 16: @Joselgc1 
If you can give me example with working expression in details of PR please

libs/shared/src/lib/components/form-modal/form-modal.component.html
Joselgc1: Implemented in new commit.

libs/shared/src/lib/utils/survey-form-action-labels.util.ts
Joselgc1: Working Expression Examples:

The new button-label settings accept either:
- Plain text, for example Save changes
- A SurveyJS expression, when the value contains {...} and resolves to a string

Examples below assume the form has a status question.

saveButtonLabel: iif({status} = 'draft', 'Continue editing', 'Save form')
Result:
- In the regular form, if status = 'draft', the button shows Continue editing
- Otherwise it shows Save form

modalSaveButtonLabel: iif({status} = 'draft', 'Update draft', 'Update record')
Result:
- In the form modal, if status = 'draft', the button shows Update draft
- Otherwise it shows Update record

saveAsDraftButtonLabel: iif({status} = 'draft', 'Update saved draft', 'Save as draft')
Result:
- If status = 'draft', the draft button shows Update saved draft
- Otherwise it shows Save as draft

Additional info:
- If the value does not contain {}, it is treated as a plain static label
- If the value contains {}, it is evaluated through survey.runExpression(...)
- The expression must resolve to a string, otherwise the raw configured value is used as fallback

One important note:
- With the current code, modalSaveButtonLabel falls back to saveButtonLabel when modalSaveButtonLabel is empty. Same for modalSaveAsDraftButtonLabel and saveAsDraftButtonLabel.

PR #2932, opened 2026-09-08
AB#137621 Fix bug with survey options rendering in Safari

Maximo-reliefapps:
# Description

In Safari, the options popup of survey `dropdown` question was rendering
correctly for a moment, then the options shifted up by a few px and stopped being
clickable.

Bad specific component styling, which was clashing with safari rendering. Moved the styling to the correct class.
Also fixed mismatch between row height and coded item height.
This fix also targets tagbox widget, which was also affected.

## Useful links

- [AB#137621](https://dev.azure.com/WHOHQ/EMSSAFE/_boards/board/t/App%20Builder%20-%20Core/Stories?System.IterationPath=@currentIteration&workitem=137621)

## Type of change

# How Has This Been Tested?

- [x] Open an option question in a survey in Safari, should display correctly
- [x] Open an option question in a survey in Chrome, should display correctly
- [x] Open a multi-option question in a survey in Safari, should display correctly
- [x] Open a multi-option question in a survey in Chrome, should display correctly

## Screenshots

Screenshot 2026-09-08 at 11 09 29

# Checklist:

( * == Mandatory ) 

# More explanation

### DIFF
--- a/libs/shared/src/lib/style/survey.scss
+++ b/libs/shared/src/lib/style/survey.scss
@@ -196,22 +196,6 @@
       }
     }
   }
-  .k-list-ul {
-    @apply shadow-lg
-      ring-black
-      ring-opacity-5
-      focus:outline-none
-      sm:text-sm
-      ease-in-out
-      duration-100
-      w-full
-      overflow-auto
-      rounded-md
-      bg-white
-      text-base
-      max-h-36
-      py-1;
-  }
   // dropdown/tagbox item
   .k-list-item {
     @apply px-2
@@ -227,7 +211,12 @@
   }
   .k-popup,
   .k-list {
-    @apply rounded-md;
+    @apply shadow-lg
+      ring-1
+      ring-black
+      ring-opacity-5
+      bg-white
+      rounded-md;
   }
 
   .sd-root-modern {
--- a/libs/shared/src/lib/survey/widgets/dropdown-widget.ts
+++ b/libs/shared/src/lib/survey/widgets/dropdown-widget.ts
@@ -165,7 +165,7 @@ export const init = (
     (question as any)._componentRef = dropdown;
     const dropdownInstance: ComboBoxComponent = dropdown.instance;
     dropdownInstance.virtual = {
-      itemHeight: 28,
+      itemHeight: 30,
     };
     dropdownInstance.valuePrimitive = Boolean(question.isPrimitiveValue);
     dropdownInstance.filterable = true;
--- a/libs/shared/src/lib/survey/widgets/tagbox-widget.ts
+++ b/libs/shared/src/lib/survey/widgets/tagbox-widget.ts
@@ -260,7 +260,7 @@ export const init = (
     (question as any)._componentRef = tagbox;
     const tagboxInstance: MultiSelectComponent = tagbox.instance;
     tagboxInstance.virtual = {
-      itemHeight: 28,
+      itemHeight: 30,
     };
     tagboxInstance.valuePrimitive = Boolean(question.isPrimitiveValue);
     tagboxInstance.filterable = true;


Maximo-reliefapps:
This is the audit I did, tested all in A & B + the big majority of the rest.
Everything is working in safari & chrome. There are some divergences in pdf previews but that's expected.


## Group A — Downloads and exports

**Highest priority.** These are engine-behaviour differences, not version-gated,
so they affect current Safari. All four were read and confirmed.

| # | Test | Where | Expected Safari divergence |
|---|---|---|---|
| A1 | Export a records grid to xlsx **and** csv; download a file from document management | [download.service.ts:288](frontend/libs/shared/src/lib/services/download/download.service.ts#L288), [document-management.service.ts:303](frontend/libs/shared/src/lib/services/document-management/document-management.service.ts#L303) | `URL.revokeObjectURL(link.href)` runs synchronously in the same tick as `link.click()`. Chrome snapshots the blob when the download starts; WebKit resolves the blob URL *after* the click is dispatched, so the download can abort or produce a 0-byte / "Failed – Network error" file. Both `saveFile` copies are identical. This covers **every** export path: `getFile`, `getRecordsExport`, `getActivitiesExport`, `getUsersExport`, and `graphqlToExcel`. |
| A2 | Export a chart widget as PNG | [chart.component.ts:381](frontend/libs/shared/src/lib/components/widgets/chart/chart.component.ts#L381) | Anchor is **never inserted into the document** before `.click()`. WebKit ignores `download` on a detached anchor; Chrome honours it. Compounded by the href being a `data:` URL, which Safari also blocks for top-level navigation. Expect a silent no-op. |
| A3 | Download a file whose content is inline (a `data:` URL rather than a storage path) | [file.service.ts:162](frontend/libs/shared/src/lib/services/file/file.service.ts#L162) | Same detached-anchor + `data:` URL pattern as A2. Note `download.service.ts` does this correctly (`document.body.append(link)` first) — the inconsistency shows A2/A3 are oversights. |
| A4 | Action button configured as **clone record → on save → navigate to URL in new tab** | [action-button.component.ts:474](frontend/libs/shared/src/lib/components/action-button/action-button.component.ts#L474) | `window.open()` is called after a GraphQL mutation resolves, i.e. outside the user-gesture stack. Chrome keeps a user-activation window; Safari's popup blocker blocks it and the post-save navigation silently doesn't happen. Other `window.open` sites are in direct click handlers — confirm them but expect them to pass. |

---

## Group B — Popups, overlays and vendor-internal CSS

Same structural family as the bug we fixed: first-party CSS applied to a
third-party library's internal layout boxes.

| # | Test | Where | What to look for |
|---|---|---|---|
| B1 | **Regression test for the fix.** Open a survey `dropdown` with a long choice list (a resource-select is ideal — >100 options exercises the virtualized path). Then a `tagbox`. Scroll to the bottom; type in the filter to drop below 100 options and scroll again | [dropdown-widget.ts](frontend/libs/shared/src/lib/survey/widgets/dropdown-widget.ts), [tagbox-widget.ts](frontend/libs/shared/src/lib/survey/widgets/tagbox-widget.ts) | Rows aligned, hover highlights the row under the cursor, every row selectable (not just the last), alignment holds to the bottom of the list and after filtering. Compare against Chrome for identical row spacing and popup height. |
| B2 | While a survey dropdown/tagbox popup is open, check its **width** | [survey.scss:55](frontend/libs/shared/src/lib/style/survey.scss#L55) — `::ng-deep .sv_row div { min-width: 0px !important }` | These two widgets are the only `appendTo: 'component'` popups in the repo, so their whole `.k-animation-container → .k-popup → .k-list → .k-list-content` chain is `<div>`s inside `.sv_row` and inherits this `!important`. Watch for a collapsed or too-narrow popup. |
| B3 | Open a form with a **PDF file-preview** question; scroll it | [survey.scss:250-293](frontend/libs/shared/src/lib/style/survey.scss#L250-L293) | Closest structural analogue to the fixed bug. First-party CSS sets `height`/`max-height`/`padding` **and flips `box-sizing` from `content-box` to `border-box`** on SurveyJS's `.sd-file__list`, which the vendor makes `position: absolute` **and** `overflow: auto`, with absolutely positioned choose/clear buttons inside. An `<iframe>` inside an `overflow: auto` box is a classic WebKit case that ignores the scroll box and expands to content height. Also: the `585px` constant is duplicated (`.sd-file` height calc vs `.file-pdf-preview__frame`) with nothing keeping them in sync. |
| B4 | Open the **email wizard**, then visit *any* Kendo TabStrip elsewhere in the app in the same session. Try a dropdown or tooltip inside a tab panel | [email.component.scss:31](frontend/libs/shared/src/lib/components/email/email.component.scss#L31), duplicated in [create-dataset.component.scss:38](frontend/libs/shared/src/lib/components/email/steps/create-dataset/create-dataset.component.scss#L38) | `overflow-x: hidden` on `.k-tabstrip-items-wrapper` and `.k-content.k-active`. Per spec, setting one axis to `hidden` makes the other compute to `auto` — so this silently turns both vendor boxes into scroll containers, clipping popups and hiding the tab-strip's own overflow arrows. Bare `::ng-deep` with no `:host`, so it leaks app-wide once either component loads. |
| B5 | Open a Monaco style editor, then trigger autocomplete in **any** Monaco instance | [custom-widget-style.component.scss:1](frontend/libs/shared/src/lib/components/custom-widget-style/custom-widget-style.component.scss#L1), [custom-style.component.scss:1](frontend/apps/back-office/src/app/components/custom-style/custom-style.component.scss#L1) | `.editor-widget.suggest-widget { left: 0 !important }` overrides the inline `left` Monaco computes, globally. Mispositioned suggest list. |
| B6 | Open a Leaflet map popup; check the tip/arrow and clickability of controls near its edges | [map-popup.component.scss:3](frontend/libs/shared/src/lib/components/ui/map/map-popup/map-popup.component.scss#L3) | `:host ::ng-deep { overflow: hidden }` (a bare declaration block, emitted onto the host) combined with `transform`-positioned tip children — a known WebKit clip/hit-test divergence. |

---

## Group C — `position: sticky`

Safari's sticky implementation diverges from Blink's in exactly these cases.

| # | Test | Where | What to look for |
|---|---|---|---|
| C1 | **Widest blast radius.** Scroll horizontally in any table with a pinned action column | ~30 templates using `[sticky]`/`[stickyEnd]` — e.g. [user-list.component.html:108](frontend/libs/shared/src/lib/views/application-users/components/user-list/user-list.component.html#L108), [notifications.component.html:59](frontend/libs/shared/src/lib/components/notifications/notifications.component.html#L59), [email.component.html:146](frontend/libs/shared/src/lib/components/email/email.component.html#L146) | Tailwind preflight sets `table { border-collapse: collapse }` (confirmed, preflight.css:151). **WebKit does not honour `position: sticky` on `<th>`/`<td>` in a collapsed-border table**; Chrome and Firefox do. Expect pinned columns to simply scroll away in Safari. The scroller itself is correct — [table-wrapper.directive.ts:36](frontend/libs/ui/src/lib/table/table-wrapper.directive.ts#L36) — so this is purely the collapse interaction. |
| C2 | Scroll the content of a **tabs widget** on a dashboard | [tabs.component.scss:11](frontend/libs/shared/src/lib/components/widgets/tabs/tabs.component.scss#L11) vs [styles.scss:86](frontend/libs/shared/src/lib/style/styles.scss#L86) | `#tab-list` is `position: sticky; top: 0` in one file, while the other gives the same id `height: 100%; overflow-y: hidden` nested inside `.widget-header`, which is itself `overflow: hidden`. A sticky element whose nearest scrollport is `overflow: hidden` never sticks, and WebKit's behaviour with `height: 100%` on the sticky element differs from Chrome's. |
| C3 | Open a `ui-select-menu` with a search box and scroll the options; compare against a `graphql-select` | [select-menu.component.html:53](frontend/libs/ui/src/lib/select-menu/select-menu.component.html#L53) vs [graphql-select.component.html:27](frontend/libs/ui/src/lib/graphql-select/graphql-select.component.html#L27) | In `select-menu` the `#search` template renders **outside** the scrolling `<ul>`, so its `sticky top-0` is inert; in `graphql-select` it is projected inside and does stick. Two visibly different behaviours from near-identical markup. Same inert pattern in [icon-picker-popup.component.html:3](frontend/libs/shared/src/lib/components/controls/icon-picker/icon-picker-popup/icon-picker-popup.component.html#L3). Also check that the panel's `overflow-hidden` isn't clipping nested tooltips. |
| C4 | Scroll inside a **TinyMCE** editor in the email layout step | [layout.component.scss:1](frontend/libs/shared/src/lib/components/email/steps/layout/layout.component.scss#L1) vs [tinymce.const.ts](frontend/libs/shared/src/lib/const/tinymce.const.ts) | `::ng-deep .tox-tinymce { height: 280px !important }` (global, no `:host`) against JS configs specifying `height: 600`/`300`/`50`. `toolbar_sticky: true` puts `position: sticky` on `.tox-editor-header`, so forcing the container height changes which box is the sticky scrollport. |

---

## Group D — Date handling

These produce **silent content differences**, not errors — the same record renders
differently in the two browsers. There is **no date library in the project**
(no moment/dayjs/date-fns/luxon); everything is hand-rolled around `new Date()`.
Safari's fallback parser rejects a much narrower set of strings than V8's.

| # | Test | Where | What to look for |
|---|---|---|---|
| D1 | Load a grid whose records contain date/datetime/time fields sourced from a **non-ISO** origin; open a record for edit | [grid.service.ts:545](frontend/libs/shared/src/lib/services/grid/grid.service.ts#L545), [core-grid.component.ts:622](frontend/libs/shared/src/lib/components/ui/core-grid/core-grid.component.ts#L622) | `new Date(value)` per cell, then `getTime() + getTimezoneOffset()` arithmetic. On Invalid Date this yields `NaN` silently rather than throwing — blank or wrong date cells in Safari, correct in Chrome. |
| D2 | Build a chart over **reference data** (external API or static CSV) using `month` / `year` / `week` aggregation | [filter.util.ts:373-425](frontend/libs/shared/src/lib/utils/reference-data/filter.util.ts#L373-L425) | The likeliest place to receive `"01/02/2024"` or `"2024-01-01 10:00"`. Chrome produces `"2024-01"`; Safari produces `"NaN-NaN"`. The surrounding `try/catch` never fires because Invalid Date doesn't throw. (Unrelated but worth noting while testing: line 405's `date.getDate() + 1` looks like a genuine off-by-one in *both* engines.) |
| D3 | Open the **query-builder date filter** on a saved filter whose stored value is a non-ISO date string | [date-filter-editor.component.ts:24-37](frontend/libs/shared/src/lib/components/query-builder/date-filter-editor/date-filter-editor.component.ts#L24-L37) | lodash `isDate()` returns `true` for an Invalid Date object, so `.toISOString()` throws `RangeError: Invalid time value` — from inside a **getter called during change detection**, meaning a repeated error every CD cycle and a dead dialog. Chrome's more permissive parsing hides this. |
| D4 | Every `uiDatePicker` field: visual alignment, the calendar icon, and a value entered near a timezone boundary (test from a negative-UTC-offset machine) | [date-picker.directive.ts:184](frontend/libs/ui/src/lib/date/date-picker.directive.ts#L184), [date-wrapper.directive.ts:150](frontend/libs/ui/src/lib/date/date-wrapper.directive.ts#L150) | Three divergences at once: (a) Safari renders `input[type=date]` with its **own** native indicator and picker on top of the app's custom Kendo overlay calendar — expect two pickers and desynced state; (b) the `format="dd/mm/yyyy"` attribute isn't real in any browser, so display order is locale-derived and differs between Chrome and Safari; (c) `min-h-[32px]`/`leading-[1.6]` don't apply to WebKit's date-field shadow parts, which need `-webkit-date-and-time-value` for vertical centering. Also check the `valueAsDate` (UTC midnight) vs local-time write-back for an off-by-one day. |
| D5 | Send an email preview containing a table with mixed string cells | [email.service.ts:1556](frontend/libs/shared/src/lib/components/email/email.service.ts#L1556), [preview.component.ts:437](frontend/libs/shared/src/lib/components/email/steps/preview/preview.component.ts#L437) | `Date.parse()` is run against **every** string cell and anything that parses is reformatted. V8 parses `"12"`, `"2025-"`, `"March"` as dates; JSC doesn't. Same content renders as a date in one browser and raw text in the other. The correct guard already exists in the codebase — [search-filters.ts:163](frontend/libs/shared/src/lib/utils/filter/search-filters.ts#L163) gates on a strict ISO regex — so use it as the model if this confirms. |
| D6 | A SurveyJS form using the `weekday()` / `addDays()` custom expression functions | [custom-functions.ts:196,280,291](frontend/libs/shared/src/lib/utils/custom-functions.ts#L196) | Inputs are fully form-builder-controlled, so arbitrary strings reach `new Date()`. Line 196's `Number(...) || ''` fallback yields `new Date('')` → Invalid Date whenever the number is `0` or `NaN`. |

---

## Group E — Version-gated

**Only run this group once the minimum supported Safari/iOS version is decided.**
If the floor is Safari 17+, all of these are moot.

| # | Feature | Safari floor | Where | Failure mode |
|---|---|---|---|---|
| E1 | Regex **lookbehind** `(?<=` | 16.4 (Mar 2023) | [context.service.ts:69,73,402,456](frontend/libs/shared/src/lib/services/context/context.service.ts#L69), [widget.service.ts:125](frontend/libs/shared/src/lib/services/widget/widget.service.ts#L125), [arcgis.service.ts:196](frontend/libs/shared/src/lib/services/map/arcgis.service.ts#L196) | **Parse-time SyntaxError**, not catchable — the whole module fails to evaluate. `ContextService` is injected in ~30 files, so below 16.4 this is a white screen, not a degraded feature. Verified: 6 occurrences, all trivially rewritable as a capture group. Lookaheads `(?=` are fine everywhere. Note the build does **not** downlevel this, so the browserslist target is irrelevant — only the real install base matters. |
| E2 | `structuredClone` | 15.4 | [edit-action-buttons-modal.component.ts:209](frontend/apps/back-office/src/app/components/edit-action-buttons-modal/edit-action-buttons-modal.component.ts#L209), [custom-row-actions.component.ts:203](frontend/libs/shared/src/lib/components/widgets/grid-settings/custom-row-actions/custom-row-actions.component.ts#L203) | Runtime `ReferenceError`; breaks only the duplicate-action-button feature. `lodash.cloneDeep` is already a dependency. |
| E3 | `addEventListener(…, { signal })` | 15 | [dropdown-widget.ts:180](frontend/libs/shared/src/lib/survey/widgets/dropdown-widget.ts#L180) | Options object silently ignored → listener never removed → one leaked handler per dropdown re-render. No error. |
| E4 | `scroll()` / `scrollIntoView()` **with options** | 15.4 | [summary-card.component.ts:649](frontend/libs/shared/src/lib/components/widgets/summary-card/summary-card.component.ts#L649), [sidenav-container.component.ts:118](frontend/libs/ui/src/lib/sidenav/sidenav-container.component.ts#L118), + 4 `scrollIntoView({behavior:'smooth'})` sites | Pre-15.4 WebKit ignores the options object entirely, so `scroll({top: 0})` is a **no-op** — the summary-card list won't reset to top on paging. Separately, [editor.service.ts:209](frontend/libs/shared/src/lib/services/editor/editor.service.ts#L209) uses the legacy boolean overload `scrollIntoView(false)`, which anchors differently in WebKit. |
| E5 | `scrollbar-gutter: stable` | 18.2 | [shared.scss:19](frontend/libs/styles/src/lib/shared/shared.scss#L19), [web-widgets/styles.scss:14](frontend/apps/web-widgets/src/styles.scss#L14) | Layout width differs between engines. |
| E6 | `vh` units on iOS | n/a — behavioural | [navbar.component.scss:80-98](frontend/libs/shared/src/lib/components/navbar/navbar.component.scss#L80-L98), [dialog.component.scss:3](frontend/libs/ui/src/lib/dialog/dialog.component.scss#L3), and 4 more | iOS Safari's dynamic toolbar makes `100vh` overflow the visible viewport. The navbar rules subtract hardcoded chrome heights (`calc(100vh - 64px - 48px - 84px)`) that must match rendered CSS. Test on a real iOS device, scrolling up and down. |

---

## Group F — `web-widgets` embed and Safari ITP

Test by embedding on a **genuinely different registrable domain**, not localhost —
that is the only way ITP's third-party rules engage.

| # | Test | Where | What to look for |
|---|---|---|---|
| F1 | Embed the widget cross-domain, authenticate, then return after **8+ days** without a top-level visit to our domain | [app.module.ts:194](frontend/apps/web-widgets/src/app/app.module.ts#L194) (`OAuthStorage` → `localStorage`), reads at [graphql.module.ts:35](frontend/apps/web-widgets/src/app/graphql.module.ts#L35), [auth-interceptor.service.ts:56](frontend/libs/shared/src/lib/services/auth-interceptor/auth-interceptor.service.ts#L56) | Safari's ITP caps script-written `localStorage` at 7 days without top-level interaction and partitions third-party storage. `idtoken`, `access_token`, `redirectPath` and `lang` all vanish. The config uses PKCE + `offline_access` (the right choice — hidden-iframe silent refresh is killed by ITP), but the refresh token lives in the same evicted storage. Confirm whether the host-mediated `refreshToken$` channel ([app-widget.component.ts:102](frontend/apps/web-widgets/src/app/widgets/app-widget/app-widget.component.ts#L102)) can fully recover. |
| F2 | Start a form-builder draft, leave it 8+ days, come back | [form-builder.component.ts:566](frontend/apps/back-office/src/app/dashboard/pages/form-builder/form-builder.component.ts#L566) | Draft recovery writes `localStorage.setItem('form:<id>', …)`. Gone in Safari, present in Chrome — users will report "Safari lost my work". Same class: summary-card page size, and the `lang`/`date-lang` keys behind the language switcher silently resetting. |
| F3 | Background the tab for 5+ minutes on macOS Safari (or sleep the device), then return | `graphql-ws` `createClient` in all four apps' `graphql.module.ts` | Safari aggressively suspends WebSockets and — unlike Chrome — often doesn't fire `close` promptly, so `graphql-ws` auto-reconnect may not trigger. Confirm subscriptions resume, and that the token in `connectionParams` is re-read on reconnect rather than captured once. |

---

